PR #11, opened 2024-07-10
chore: add initial ci

TristanHoladay:
Setting up some initial CI and issue templates


.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - id: detect-aws-credentials
+        args:
+          - "--allow-missing-credentials"
+      - id: detect-private-key
+      - id: end-of-file-fixer
+        exclude: docs/command-reference
+      - id: fix-byte-order-marker
+      - id: trailing-whitespace
+        args: [--markdown-linebreak-ext=md]
+  - repo: https://github.com/sirosen/fix-smartquotes
```
jeff-mccoy, line 17: this is a deprecated repo, that hasnt been touched in 4 yrs, also generally not a fan of brining ing random code like this

tasks/test.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  - name: e2e
+    actions:
+      - task: setup:node_modules
+      - cmd: which playwright || npx playwright install
+      - task: setup:slim-cluster
+      - cmd: npm run test:integration
+        dir: ui
+
+  - name: stream
```
jeff-mccoy, line 19: not sure i follow the need to break these tests out by package?

tasks/setup.yaml
```diff
@@ -0,0 +1,27 @@
...
+    description: "Install golangci-lint to GOPATH using install.sh"
+    actions:
+      - cmd: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.59.1
+
+  - name: node_modules
+    description: "Install node_modules if not present"
+    actions:
+      - cmd: |
+          if [ ! -d "node_modules" ]; then
+            npm install
```
jeff-mccoy, line 25: we should never use `npm i` like this, only `npm ci`, not sure I love the idea of just seeing if node_modules exists, locally this could cause all sorts of interesting problems

jeff-mccoy (changes requested):
mostly great, just a fe clarifying questions

.pre-commit-config.yaml
TristanHoladay: good call out. we could look to change this in CLI then as well. Probably don't even need it honestly. Looks like[ Zarf ](https://github.com/defenseunicorns/zarf/blob/eca0c530a04c3ed155c03761eb45bd16ba717b79/.pre-commit-config.yaml#L19 ) is using a [newer repo](https://github.com/sirosen/texthooks/blob/9494d633234f89a66d3429f692bc9715e37db1f6/.pre-commit-hooks.yaml#L1) of his to call the same hook.

tasks/test.yaml
TristanHoladay: yeah i didn't like that either, I was just having trouble running them together at the time so did that and moved on. but i'll fix that.

.github/actions/k3d/action.yaml
```diff
@@ -0,0 +1,11 @@
+name: setup-k3d
+description: "Install k3d and create a cluster"
+
+runs:
+  using: composite
+  steps:
+    - run: "curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash"
+      shell: bash
+
+    - run: k3d cluster delete && k3d cluster create --image="rancher/k3s:v1.28.4-k3s2"
```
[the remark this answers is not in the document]
TristanHoladay: oh yeah ... huh, i didn't look at that close enough then. i thought it was just installing k3d. i'll clean that up since we're using slim-dev for e2e anyway.

.github/workflows/scorecard.yaml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@dc50aa9510b46c811795eb24b2f1ba02a914e534 # v2.3.3
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          repo_token: ${{ secrets.SCORECARD_READ_TOKEN }}
```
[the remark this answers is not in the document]
TristanHoladay: ahh yes, we do need that please.

tasks/setup.yaml
jeff-mccoy: in ci none--but if local, you could have an out of date node_modules, think pull down main and there were news deps added or changed, checking will break that

TristanHoladay: oh yeah, duh. probably best to just `ci` each time then. shouldn't be that big of an add to the test execution time.

jeff-mccoy (approved):
awesome job knocking out all this CI setup 🦄

Issue #23, opened 2024-07-19
Export monitor pepr data to JSON


UncleGedd:
### Is your feature request related to a problem? Please describe

Add an action to the Export button on the Monitor/Pepr page to export the data to a JSON file
![Screenshot 2024-07-19 at 12 51 12 PM](https://github.com/user-attachments/assets/fbc12258-641f-4298-b98d-32831727ab5f)

PR #24, opened 2024-07-19
feat: export pepr data stream

UncleGedd:
## Description

- adds feature to export monitor/pepr data as a JSON file
- adds typescript section pre-commit
- adds import optimizer prettier plugin (specifically added to remove unused imports)
- adds Maru task for builds

## Related Issue

- #23


jeff-mccoy (approved):
good add & good to normalize import formatting

PR #28, opened 2024-07-23
feat: add sparse streams, compression, single resource api endpoints

jeff-mccoy:
## Description

The original SSE implementation is very inefficient as it streams the complete / reconciled list on each data change event. After exploring a number of different strategies for optimizing this, including differentials, sparse data + compression realized a significant efficiency gains, > 90% typically and in some extreme cases up to 99.96% reduction in bandwidth requirements.

Spare Streams: This PR introduces a concept of sparse and dense resource streams. The default is now sparse and can be overriden by adding the query param `dense=true` to any of the resource streams. Sparse streams only return `metadata`, `status`, `type`, `data` (keys only), `kind`, `apiVersion` fields and strip `managedFields` from `metadata` as well. This results in much lighter data streams.

Single-Resource Endpoint: This PR also introduces the ability to call a dense resource by it's UID, to be used in conjunction with thew new spare streams. The url pattern is `/api/v1/resources/<kind>/uid`, in this mode `once` and `dense` currently have no affect, this also means we do not yet support streaming a single resource.

Compression: GZIP compressions for both streams and regular API calls will now automatically be enabled if the client supports it.

This PR also introduces the Secrets & Configmaps UI views, which originally triggered the performance concerns due to large binary data stored in their resource lists.

Sample data on a basic [UDS Core Demo](https://github.com/defenseunicorns/uds-core?tab=readme-ov-file#quickstart) deployment for the `pods`:

| Path                                                  | Sparse | Dense | Compressed |    Size |
|-------------------------------------------------------|:------:|:-----:|:----------:|--------:|
| /api/v1/resources/workloads/pods?once=true&dense=true |        |   ✅   |            |  477 kB |
| /api/v1/resources/workloads/pods?once=true            |    ✅   |       |            |  168 kB |
| /api/v1/resources/workloads/pods/5b4bbc74-...         |        |   ✅   |            | 10.3 kB |
| /api/v1/resources/workloads/pods?once=true&dense=true |        |   ✅   |      ✅     | 46.1 kB |
| /api/v1/resources/workloads/pods?once=true            |    ✅   |       |      ✅     | 22.3 kB |
| /api/v1/resources/workloads/pods/5b4bbc74-...         |        |   ✅   |      ✅     |  3.3 kB |


BillyFigueroa:
One of the CI checks will be fixed by running `npm run format` inside the UI folder. There are some formatting errors the linter is catching that will be fixed running that

UncleGedd:
Testing the the `uid` URL param. I may be doing something wrong, the `;uid` of this deployment resource is returning a 404 in the UI, but it's definitely in the cluster
Screenshot 2024-07-24 at 8 24 05 AM


Screenshot 2024-07-24 at 8 25 40 AM

pkg/api/sse/bind.go
```diff
@@ -8,35 +8,81 @@ import (
...
+		// By default, send the data as a sparse stream
+		once := r.URL.Query().Get("once") == "true"
+		dense := r.URL.Query().Get("dense") == "true"
 
-			// Set the headers
-			w.Header().Set("Content-Type", "text/json; charset=utf-8")
-			w.Header().Set("Cache-Control", "no-cache")
-			w.Header().Set("Connection", "keep-alive")
+		// Get the UID from the URL if it exists
+		uid := chi.URLParam(r, "uid")
```
UncleGedd, line 23: Playing around with the `uid` param and I can't seem to get it populated. I'm hitting:
```
http://localhost:8080/api/v1/resources/workloads/deployments?uid=6ef2affd-8c52-43fc-89e1-57f6189e205a
```

But the `uid` var is an empty string
Screenshot 2024-07-24 at 8 37 12 AM

pkg/api/sse/bind.go
```diff
@@ -8,35 +8,81 @@ import (
...
 
-			// Write the data to the response
-			if _, err := w.Write(data); err != nil {
-				http.Error(w, "Failed to write data", http.StatusInternalServerError)
+		// If a UID is provided, send the data for that UID
+		// Streaming is not supported for single resources
+		if uid != "" {
```
UncleGedd, line 27: Is there any place in the frontend where we're using this param or this a "later" sort of thing?

UncleGedd:
Any ideas for a backend testing strategy?

pkg/api/sse/bind.go
UncleGedd: nvm, I was mixing up query params vs url params. This works: 

```
http://localhost:8080/api/v1/resources/workloads/deployments/6ef2affd-8c52-43fc-89e1-57f6189e205a
```

pkg/api/sse/bind.go
jeff-mccoy: Need this for the details views #34

jeff-mccoy:
> Any ideas for a backend testing strategy?

We need to decide/create the unit tests and probably separate API tests as well for this. TBD, but should prioritize soon.

TristanHoladay (approved):

UncleGedd (approved):

decleaver (approved):
🚀

PR #67, opened 2024-07-24
fix(ui): pepr monitor unsubscriber

### LINKED ISSUE #37: pepr monitor view not closing properly
Jeff found that pepr is not closing properly creating multiple endpoints. 

As you migrate from one view to another from pepr monitor the old pepr is not removed creating multiple.

decleaver:
## Description

fixes bug where we were getting an additional pepr subscriber everytime we would navigate away from the pepr view and back.

fixes #37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ vite.config.ts.timestamp-*
 
 # # Playwright/ Vitest
 /ui/test-results
+
+# Dev
+tmp/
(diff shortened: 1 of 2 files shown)


TristanHoladay (approved):

jeff-mccoy (approved):

jeff-mccoy (approved):
tested locally, lgtm

Issue #30, opened 2024-07-23
Test Stores


TristanHoladay:
### Describe what should be investigated or refactored
We need to develop a pattern for testing the ui stores. Stores are found under each feature in `/ui/src/lib/features`.

adam-defenseunicorns:
Jeff "make an issue for adding a testing pattern + test for all the store.ts files used by the k8s resource cores in the ui"

PR #74, opened 2024-07-25
chore: add resource store tests

TristanHoladay:
## Description
Add pattern and tests for the k8s resource stores

## Related Issue

- #30


TristanHoladay:
For testing the pod store it became too complicated to use the test helper, so opted for duplicating the mocking logic and then modifying to handle the more advanced cases.

UncleGedd (approved):
🚀

PR #75, opened 2024-07-25
feat(ui): networkpolicies view

decleaver:
## Description
 UDS UI:

image

Lens:
image


jeff-mccoy (changes requested):
K9s has a better view here I think we should implement for this.

Screenshot 2024-07-26 at 2 23 01 AM

decleaver:
updated view matching k9s:
image

jeff-mccoy (approved):

PR #82, opened 2024-07-25
feat: pvc with pods

UncleGedd:
## Description

Adds pods to PVC view to match Lens

Lens:
Screenshot 2024-07-26 at 3 47 36 PM

UDS:
Screenshot 2024-07-26 at 3 47 59 PM


UncleGedd:
blocked by https://github.com/defenseunicorns/uds-runtime/issues/85

Madeline-UX (approved):
Looks good from UX perspective. 

I need to provide guidance on status colors but that should not block

UncleGedd:
> Not ignoring this one, just waffling on the fact we have to bring in dense stream of pods and it makes me very sad.....

np, happy to refactor after we implement field selectors

UncleGedd:
blocked by https://github.com/defenseunicorns/uds-runtime/pull/94

Going to refactor to use selectors instead of pulling dense pod data

Issue #89, opened 2024-07-26
Setup deployment of UDS Runtime


UncleGedd:
### Is your feature request related to a problem? Please describe

Need to provide a mechanism by which users can deploy UDS Runtime.
- Add Dockerfile
- Add zarf.yaml
- Create a UDS Package out of the runtime?

PR #93, opened 2024-07-27
feat: enables uds-runtime releases

UncleGedd:
## Description

Adds everything necessary to deploy (besides CI, will add that in a separate PR, see https://github.com/defenseunicorns/uds-runtime/issues/96)

Publish with: `uds run build:publish-uds-runtime` 

uds-runtime image: `ghcr.io/defenseunicorns/uds-runtime:alpha`
uds-runtime Zarf pkg: `ghcr.io/defenseunicorns/packages/uds/uds-runtime:alpha`

Adds
- Dockerfile
- Helm chart including UDS Package
- load testing script (used to configure resource limits in chart)
- `zarf.yaml`
- tasks to support publishing all the things

Resolves https://github.com/defenseunicorns/uds-runtime/issues/89


chart/values.yaml
```diff
@@ -0,0 +1,5 @@
+replicaCount: 1
+image:
+  repository: ghcr.io/unclegedd/uds-runtime-dev
```
jeff-mccoy, line 3: will have to update this when ready

zarf.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: uds-runtime
+  description: tbd
+  version: 0.1.0
+
+components:
+  - name: uds-runtime
+    required: true
+    images:
+      - ghcr.io/unclegedd/uds-runtime-dev:dev
```
jeff-mccoy, line 11: same with this one

Dockerfile
```diff
@@ -0,0 +1,40 @@
+# Stage 1: Build Svelte frontend
+FROM node:21-alpine AS frontend-builder
```
jeff-mccoy, line 2: can we switch all of these to chainguard images?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+
+# Stage 2: Build Go backend for amd64
+FROM golang:1.22.4-alpine AS backend-builder-amd64
+WORKDIR /app
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+COPY --from=frontend-builder /app/ui/build ./ui/build
+RUN CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -o /app/main-amd64 .
```
jeff-mccoy, line 15: I dont think we need to build these parts inline, we can probably let our normal build stages create the binary and then set an arch target like we do with [zarf](https://github.com/zarf-dev/zarf/blob/main/Dockerfile)

Dockerfile
UncleGedd: done

Dockerfile
UncleGedd: 👍 set up tasks to do the build and then Dockerfile picks it up

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - containerPort: 8080
+          resources:
+            requests:
+              memory: "128Mi"
+              cpu: "250m"
+            limits:
+              memory: "256Mi"
```
jeff-mccoy, line 28: Not sure how much data we have on resource usage yet, maybe we should bump the limits higher until we get that data?

chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,25 @@
...
+        gateway: tenant
+        port: 8080
+        targetPort: 8080
+    allow:
+      - direction: Egress
+        selector:
+          app: uds-runtime
+        remoteGenerated: KubeAPI
+
+      - direction: Ingress
```
jeff-mccoy, line 22: Do we need ingress here?

zarf.yaml
```diff
@@ -0,0 +1,20 @@
+kind: ZarfPackageConfig
+metadata:
+  name: uds-runtime
+  description: tbd
+  version: '###ZARF_PKG_TMPL_REF###'
```
jeff-mccoy, line 5: We should be using release please or similar like uds code does. Package templates will eventually die.

jeff-mccoy (changes requested):
Some comments

chart/templates/deployment.yaml
UncleGedd: 👍 bumping limits

chart/templates/uds-package.yaml
UncleGedd: good catch, removing

zarf.yaml
UncleGedd: updated to use release please

jeff-mccoy (approved):

PR #94, opened 2024-07-28
feat(api): add fields query parameter for fine-grained data filtering

jeff-mccoy:
# Add 'fields' query parameter for path filtering

## Overview

This PR introduces a new `fields` query parameter that enables path filtering to limit the returned data in API responses. This feature complements the existing sparse and dense response options, providing fine-grained control over the returned data structure. This feature is particularly valuable for complex or large resources where tailored data retrieval can significantly reduce bandwidth requirements.

## Key Features

- New `fields` query parameter for fine-grained control over returned data
- Supports dot notation for nested fields and array access
- Preserves the overall structure of the response
- Applies to both streaming and non-streaming responses
- Works with both list and single resource requests
- Provides an efficient middle ground between sparse and dense responses

## Implementation Details

The core of this feature is implemented in the `rest` package, primarily in the `jsonMarshal` and `filterItemsByFields` functions. These functions work together to process the `fields` parameter and filter the response data accordingly.

Key components:

1. `jsonMarshal`: Handles the initial processing of the payload and fields list
2. `filterItemsByFields`: Filters items based on the specified field paths
3. `getNestedValueFromUnstructured`: Retrieves nested values from unstructured data
4. `setNestedValue`: Sets nested values in a map structure

## Response Types and When to Use `fields`

1. **Sparse Response (Default)**: 
   - Includes only `metadata`, `status`, and a few other select fields
   - Suitable for quick listings and basic information retrieval

2. **Dense Response** (`dense=true`):
   - Provides the complete resource with all fields
   - Useful when all data for a resource is needed

3. **Fields-Filtered Response** (new `fields` parameter):
   - Allows selection of specific fields, including nested ones
   - Best used for complex or large datasets where more than the sparse data is needed, but less than the full dense response
   - Provides a balance between data completeness and efficiency

The `fields` parameter is most beneficial when:
- The resource is large or complex
- Only a specific subset of the data is needed
- The required data is more than what's included in the sparse response, but less than the full dense response

For smaller resources, using `dense=true` might be simpler if most of the data is needed.

### API Request

```
GET /api/v1/resources/workloads/pods?fields=.metadata.name,.spec.volumes[].persistentVolumeClaim.claimName,.status.phase&once=true
```

### Response (truncated)

```json
[
  {
    "metadata": {
      "name": "promtail-77jqr"
    },
    "status": {
      "phase": "Running"
    }
  },
  {
    "metadata": {
      "name": "istiod-5c6f7f8c67-q4zwp"
    },
    "status": {
      "phase": "Running"
    }
  },
  {
    "metadata": {
      "name": "loki-write-1"
    },
    "spec": {
      "volumes": [
        {
          "persistentVolumeClaim": {
            "claimName": "data-loki-write-1"
          }
        }
      ]
    },
    "status": {
      "phase": "Running"
    }
  }
]
```

## Performance Impact

The `fields` parameter allows for more efficient data transfer compared to dense responses, especially for large resources. It provides a middle ground between sparse and dense responses, optimizing both performance and data completeness based on specific needs.

## Swagger Documentation

The `fields` parameter has been added to the Swagger documentation for affected endpoints. Here's an example of the updated Swagger binding:

```go
// @Param fields query string false "Filter by fields. Format: .metadata.labels.app,.metadata.name,.spec.containers[].name,.status"
func getCronJob(cache *resources.Cache) func(w http.ResponseWriter, r *http.Request) {
	return rest.Bind(cache.CronJobs)
}
```

## Testing

- Unit tests have been added for the new filtering functions
- Integration tests have been updated to include scenarios with the `fields` parameter
- Manual testing has been performed on various resource types and field combinations
- Comparative tests between sparse, dense, and fields-filtered responses have been conducted to verify efficiency gains


jeff-mccoy:
Will pick this back up in a couple weeks, have a major refactor WIP locally I need to push up

BillyFigueroa (dismissed):

pkg/api/rest/filter.go
```diff
@@ -0,0 +1,243 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// jsonMarshal marshals the payload to JSON and filters the fields if specified
```
UncleGedd, line 14: I was able to refactor this code using client-go's `jsonpath` pkg over here: https://github.com/defenseunicorns/uds-runtime/pull/249

Ends up being well over 100 LOC shorter

UncleGedd: Disregard, I'll make a ticket to refactor later if necessary

UncleGedd (approved):
Awesome. Tested with a few calls and everything works as expected. I think we can prob simplify the `filter.go` algos using client-go's `jsonpath` pkg, but happy to make a ticket and work that later

jeff-mccoy:
Yeah i rewrote it using jsonpath from k8s but it has some major shortcomings with union types, caused me a ton of headaches so I switched back to this

Issue #34, opened 2024-07-23
K8s Resource Detail View


TristanHoladay:
### User Story

As a user of the platform management tool,
I want to be able to click on any element within the application
So that a right drawer appears with detailed information about the selected element

### Acceptance Criteria

- **Given** I am an SRE, using the UDS web app
- **When** I click on a resource/element in the app
- **Then** I see a view with details of that resource

A drawer slides in from the right side of the screen.
The drawer contains multiple sections with relevant information about the selected element.
Each section can display multiple lines of data.

### Additional context
The data in the drawer can include:

Text
Links (clickable and visually distinct)
Icons (relevant to the information being presented)

The drawer should have a clear way to close it (e.g., a close button or clicking outside the drawer).
The drawer should not obstruct critical parts of the main interface.
The transition of the drawer opening and closing should be smooth and animated.

How might we ensure information is both scannable and readable when dene?

## Designs
[Figma Dev Mode Link](https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=3664-12791&m=dev)

Component: Drawer
Placement = Right
Backdrop = False
body scrolling = true
edge = false

<img width="1434" alt="Screenshot 2024-07-26 at 11 19 22 PM" src="https://github.com/user-attachments/assets/882b5148-2836-4eef-b935-28a458e172a9">

<img width="712" alt="Screenshot 2024-07-26 at 11 24 19 PM" src="https://github.com/user-attachments/assets/7b03e503-475e-4d77-9512-123d69cb333a">

Madeline-UX:
Examples from Lense
<img width="1028" alt="Screenshot 2024-07-25 at 11 46 15 AM" src="https://github.com/user-attachments/assets/43cdaf07-1eb4-4fbb-a714-3ee424bb2fe9">

PR #95, opened 2024-07-29
feat(ui): resource detail view

jeff-mccoy:
## Description

This PR introduces K8s detail views via a drawer component. The view is addressable as it also uses a route param, e.g.: `/workloads/pods/f49541dd-b571-49bd-9ce6-2824e015020a`. In addition to showing basic metadata, the raw yaml (un-styled currently) is also shown in a tabbed interface. Keyboard navigation is also provided:

- Up/Down: switching between resources in the table
- Left/Right: switching between tags: Metadata, YAML, Events
- Esc: close the details view

## Related Issue

- #34 

Screenshot 2024-08-01 at 3 27 47 AM

Screenshot 2024-08-01 at 3 31 48 AM


ui/src/lib/components/DataTable/component.ts
```diff
@@ -0,0 +1,8 @@
+import type { KubernetesObject } from '@kubernetes/client-node'
+
+export async function loadResourceDetail(url: string, uid: string) {
```
TristanHoladay, line 3: broader question but this made me think of it. do we have some kind of error handling for network calls in place yet and if not do we want that atm?

jeff-mccoy: great question--we should add an issue for that, it should be addressed for sure

decleaver:
created follow on issue: https://github.com/defenseunicorns/uds-runtime/issues/117

decleaver (approved):

BillyFigueroa:
@jeff-mccoy do we want to keep the naming structure from flowbite to figma for this component? We have been using the term Drawer all along. Everything would be aligned if this was `components/Drawer/component.svelte`

Screenshot 2024-07-31 at 4 20 58 PM

Screenshot 2024-07-31 at 4 22 01 PM

jeff-mccoy:
Sure we can rename it drawer

UncleGedd:
Pulled it down, everything seems to work really well. I especially like the keyboard events

Were you able to use any Flowbite blocks to build this?

Madeline-UX (approved):
Passed UX/UI review

Issue #84, opened 2024-07-26
Data Table header and footer update


adam-defenseunicorns:
### Is your feature request related to a problem? Please describe
Currently, our datatable doesn't display:
- the total number of items it contains
- Table name
This lack of information makes it difficult for users to quickly understand the scale of the data they're working with. 

### Acceptance Criteria
We need to add a feature that displays the total number of items in the datatable:

- **Given** a datatable with a set of items
- **When** the datatable is loaded or updated
- **Then** the total number of items should be clearly displayed above the table

- **Given** a datatable with a set of items
- **When** the user filters or search the table 
- **Then** the number of items showing is added with the total number of items shown in parenthesis
- **And** the total number of items remains the constant at the bottom footter of the the table.

Specific requirements:
1. The total count should be updated dynamically if items are added or removed
2. The display should be clear and unobtrusive, possibly in the format "showing x of x"
3. There should be a total number of items in the footer
4. No pagination is used at this time.

## Design 
[Figma Dev Mode link](https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=3654-35116&m=dev&t=mgowQxoayQJvsUcl-1)

Flowbite Block - [Application UI Advanced Table](https://flowbite.com/blocks/application/advanced-tables/)

- [ ] Heable header = two level with search
    - [ ] button on first level > export 
    - [ ] hide primary bottom
    - [ ]  hide actions select button
    - [ ] Change search to include name of object in table (ex. Pods table > search pods)
- [ ] Table footer = button and text
    - [ ] Hide footer button

![Screenshot 2024-07-26 at 2 37 31 PM](https://github.com/user-attachments/assets/db9377d9-e0b6-4f5a-bd11-ee14127a9658)

<img width="986" alt="Screenshot 2024-07-26 at 2 21 06 PM" src="https://github.com/user-attachments/assets/98bcd9c5-2b8b-43b3-924a-872f0f6a8f87">

### Additional context (openlens)
![image](https://github.com/user-attachments/assets/9709d8b6-53fa-4227-8825-404e6ce06939)

PR #106, opened 2024-07-30
feat(ui): add header to tables

UncleGedd:
## Description

Implements Datatable header

## Related Issue

- #84


ui/src/lib/components/DataTable/component.svelte
```diff
@@ -25,6 +28,24 @@
   const rows = createStore()
   const { namespace, search, searchBy, searchTypes, sortAsc, sortBy } = rows
 
+  // get the title from the route
+  const flatRoutes = routes.flatMap((route) => {
```
UncleGedd, line 32: Nobody look at this....it's super jank and tightly couples the view to the routes (we do this with Breadcrumbs but this feels nasty here)

Going to refactor and pass title + tooltipDesc as props to DataTable

jeff-mccoy: image

ui/src/lib/components/DataTable/component.svelte
```diff
@@ -33,6 +36,28 @@
...
   <div class="table-container">
     <div class="table-content">
+      <div class="table-header">
+        <h5 class="flex items-center">
+          <span class="dark:text-white">{name}&nbsp;&nbsp;</span>
+          <span class="text-gray-500">
+            {$rows.length} / {$numResources} results
+          </span>
+          <InformationFilled
+            data-tooltip-target="resource-tooltip"
```
TristanHoladay, line 46: not entirely sure, but you might not need these data attributes. in `ui/src/lib/features/k8s/configs/uds-exemptions/exemption-details/component.svelte` there's an example of  tooltip without them.

jeff-mccoy: Yep those data attributes were for the flowbite js, but our tooltip is pure css if you have the right classes

ui/src/lib/utils/descriptions.ts
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024-Present The UDS Authors
+
+// Generated by robots and the K8s docs
+export const resourceDescriptions: { [key: string]: string } = {
+  ClusterRoleBindings:
```
TristanHoladay, line 6: this wouldn't work for airgapped, but for connected environments would it be cool to have a link to each resource in the K8s docs?

UncleGedd: Good be a nice feature down the line, we'll defer to the air gap use case for now

Madeline-UX (approved):
I sanction and approve these changes 
- change info icon to not filled
- change number to say "showing x of total"

ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -23,8 +23,17 @@
...
+  let { name, desc } = nameAndDesc
+
   const rows = createStore()
-  const { namespace, search, searchBy, searchTypes, sortAsc, sortBy } = rows
+  const { namespace, search, searchBy, searchTypes, sortAsc, sortBy, numResources } = rows
+
+  // check for filtering
+  let isFiltering = false
+  $: {
+    isFiltering = !!($search || $searchBy != 'Anywhere' || $namespace)
```
jeff-mccoy, line 35: I wouldn't include searchby as it only has meaning if search has contents

ui/src/lib/features/k8s/cluster-ops/hpa/component.svelte
```diff
@@ -15,6 +17,12 @@
     ['age'],
     ['status'],
   ]
+
+  const resourceName = 'HorizontalPodAutoscalers'
+  const nameAndDesc: NameAndDesc = {
```
jeff-mccoy, line 22: Placeholder comment to reference in PR review

jeff-mccoy (changes requested):
I left a placeholder comment to see what I'm referring to, but this is every one of those files so bringing up to here instead:

--

This is preference--but feel this would be cleaner as two unique properties and set directly in this file where all presentation parts live for this view, something like 


```
  const name = "HorizontalPodAutoscaler"
  const description = "..."
```

For testing I'm not sure we really need to test every word of the description either vs just the string exists 🤔 but not sure on that. 

Realize that's different than how we did in [kfc](https://github.com/defenseunicorns/kubernetes-fluent-client/blob/main/src/kinds.ts), but think makes more sense in the same file for this project. 

I also wouldn't be object to skipping variable declaration and putting on the component prop directly if it was still readable.

UncleGedd:
> For testing I'm not sure we really need to test every word of the description either vs just the string exists 🤔 but not sure on that.

The goal of the testing  this was to ensure the descriptions matched up to something in `descriptions.ts` (esp since it was generated by 🤖 )

ui/src/lib/components/k8s/DataTable/component.svelte
UncleGedd: removed

ui/src/lib/features/k8s/cluster-ops/hpa/component.svelte
UncleGedd: updated!

jeff-mccoy (approved):

PR #108, opened 2024-07-30
feat(ui): adding row components

BillyFigueroa:
## Description
Creating row component

## Related Issue

- #


ui/package.json
```diff
@@ -27,11 +27,13 @@
     "autoprefixer": "10.4.19",
     "carbon-icons-svelte": "12.10.0",
     "date-fns": "3.6.0",
+    "dompurify": "^3.1.6",
     "eslint": "9.7.0",
     "eslint-config-prettier": "9.1.0",
     "eslint-plugin-svelte": "2.43.0",
     "flowbite": "2.4.1",
     "globals": "15.8.0",
+    "highlight.js": "^11.10.0",
```
jeff-mccoy, line 36: Can you changes those to pinned versions, no ^ (slsa recommendation, we use ci to manage updates)

ui/src/lib/components/index.ts
```diff
@@ -2,5 +2,7 @@
 // SPDX-FileCopyrightText: 2024-Present The UDS Authors
 
 export { default as AnsiDisplay } from './AnsiDisplay/component.svelte'
+export { default as Row } from './Drawer/Row.svelte'
```
jeff-mccoy, line 5: Are these both used externally? If the components are only used in drawer, for example, I'd put them under that folder and not re-export them

jeff-mccoy (changes requested):
Few more comments

jeff-mccoy:
going to close this for now until we have more data on need for UI rendering of resource data

PR #114, opened 2024-07-31
feat: swagger

decleaver:
## Description

Adds swagger 
image

image

## Details
Using: https://github.com/swaggo

Swaggo needs the api operation comments on each handler function, so I did need to add wrappers around the sse.Bind calls, and i added all of those in swagger.go

Good News:
- All the endpoints seem to work as expected
Bad News:
- Swagger doesn't seem to support text/event stream (so `once=true` must be set otherwise it just hangs)
  -  can probably write some scripts to handle this but this would take a lot of customizing and we would probably lose autogen capabilities.
- Swaggo currently doesnt support OpenApi v3 (currently being "slowly" in their v2 branch)
   - most of the suggestions I've seen for goloang and OpenApi v3  say to use [huma](https://huma.rocks/) rest api framework 😞 
   
 I think this is ok for now, but we will want to revisit longterm.


UncleGedd (approved):
I think this is a great starting point, probably some more optimizations to be made to reduce boilerplate (I'm surprised you have to manually type the route in those annotations). IMO this is g2g 🚀

jeff-mccoy (approved):

Issue #112, opened 2024-07-31
Stylized and Colorized YAML View


adam-defenseunicorns:
# Feature Request: Stylized and Colorized YAML

## Is your feature request related to a problem? Please describe

YAML files are difficult to read and parse visually, especially for complex configurations. Users frequently struggle to understand the structure and hierarchy of YAML data, leading to errors in configuration and increased cognitive load when working with YAML files.

## Describe the solution you'd like

- **Given** a YAML file or YAML-formatted string input
- **When** the application renders the YAML content
- **Then** it should display a stylized and colorized view of the YAML data, with the following features:
  1. Syntax highlighting for different YAML elements (keys, values, arrays, etc.)
  2. Indentation guides to clearly show hierarchy
  3. Wrap text
  4. Parity with how UDS is doing YAML

## Additional context

This feature would greatly improve the user experience when working with YAML configurations, especially in applications that heavily rely on YAML for settings, deployments, or data exchange. It would reduce errors caused by misreading YAML structure and make it easier for both novice and experienced users to work with complex YAML files.

A mockup or prototype of the desired interface would be helpful to illustrate the concept. Additionally, performance considerations should be taken into account for rendering large YAML files without significant lag.

adam-defenseunicorns:
![image](https://github.com/user-attachments/assets/b0634909-d3cf-4bb3-b6eb-2d5fa24013a2)

PR #121, opened 2024-08-01
feat(ui): 112 stylized and colorized yaml view

BillyFigueroa:
## Description
Converting JSON data to YAML data in order to display useful information to users and also styling the YAML to match the other terminal tools theme

## Related Issue

- #112 

## Screenshots
Screenshot 2024-08-01 at 11 43 52 AM


ui/src/lib/features/toast/store.ts
```diff
@@ -13,8 +13,6 @@ export type Toast = {
 export const toast = writable<Toast[]>([])
 
 export const addToast = (newToast: Toast) => {
-  console.log('addToast', newToast)
```
jeff-mccoy, line 16: I actually left this log here intentionally since the error messages auto close in the ui

ui/src/lib/components/k8s/Drawer/component.svelte
```diff
@@ -2,19 +2,26 @@
...
   import { onMount } from 'svelte'
+  import * as YAML from 'yaml'
+  import hljs from 'highlight.js/lib/core'
+  import yaml from 'highlight.js/lib/languages/yaml'
+  import DOMPurify from 'dompurify'
 
   import { goto } from '$app/navigation'
   import './styles.postcss'
 
-  export let resource: KubernetesObject
```
jeff-mccoy, line 12: Was there a reason to change this? We will have a lot more types then just pods

ui/src/lib/features/toast/store.ts
BillyFigueroa: > I actually left this log here intentionally since the error messages auto close in the ui

I ll add it back

ui/src/lib/components/k8s/Drawer/component.svelte
BillyFigueroa: > Was there a reason to change this? We will have a lot more types then just pods

Changed it back

jeff-mccoy (approved):

PR #132, opened 2024-08-01
chore(ui): remove breadcrumb

jeff-mccoy:
Per @Madeline-UX we are no longer going to include the breadcrumb nav element in the UI

Screenshot 2024-08-01 at 9 20 12 PM


Madeline-UX (changes requested):
Top apdding tailwinds 10 - 40px
Screenshot 2024-08-01 at 5 46 02 PM

PR #157, opened 2024-08-06
chore: add test infra for dummy cluster

### LINKED ISSUE #148: Ephemeral Test Cluster
The team needs an easily accessible, demo environment for testing and demonstration purposes without requiring installation or complex setup. Currently, there's no simple way for people to quickly interact with and evaluate UDS Runtime.

**Given** a need for a demo environment (cluster)
**When** a demo cluster is set up with UDS Runtime
**Then** it should:

- Be accessible to Defense Unicorns
- Be ephemeral and recreated regularly (e.g., every hour, six hours, or day)
- Have minimal security measures due to its temporary nature
- Be easily shareable via a link 
- Possibly use an AWS load balancer URL with hashes
- Have no access to perform significant actions
- Serve as a continuously available "snapshot" for immediate usability testing and idea validation

Integrating into staging environments or UDS core was considered but deemed to require more careful security considerations. (Future iteration for longer term testing)

### Additional context
- While secrets may be exposed, it's not a major concern due to the cluster's ephemeral nature and limited access.
- The primary goal is to provide a simple, accessible way for people to interact with and test the system quickly.

TristanHoladay:
## Description
Add the infra, bundle, and tasks for deploying runtime with core in an ephemeral testing cluster on aws.

## Related Issue

- Resolves #148


.github/test-infra/terraform/setup.sh
```diff
@@ -0,0 +1,32 @@
...
+# Define the file path
+ file_path="uds-k3d/chart/templates/nginx.yaml"
+
+# # Replace 'uds.dev' with 'exploding.boats'
+ sed -i 's/uds\.dev/burning.boats/g' "$file_path"
+
+# # Deploy cluster
+cd uds-k3d && uds run
+
+# Get kubeconfig for uds to find
```
decleaver, line 15: I know we looked at this, still seems weird that you have to do this, but at least we know it works 🤷‍♂️

.github/test-infra/terraform/main.tf
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Attach EIP to Instance
+resource "aws_eip_association" "runtime" {
+  instance_id   = aws_instance.runtime.id
+  allocation_id = data.aws_eip.runtime_eip.id
+}
+resource "aws_iam_instance_profile" "runtime_profile" {
+  name = "runtime-ephemeral-EC2InstanceProfile"
+  role = aws_iam_role.runtime_role.name
```
UncleGedd, line 65: Do we need to add tags to any of these other resources? ie.
- aws_eip_association
- aws_iam_instance_profile
- ssm_parameter_policy

.github/test-infra/terraform/setup.sh
TristanHoladay: yeah just tested without it that and it fails

.github/test-infra/terraform/main.tf
TristanHoladay: good catch. i tagged everything that would let me. the policy is in a weird state atm that i can't tag it because of permissions. @joelmccoy is tracking.

.github/test-infra/terraform/main.tf
```diff
@@ -0,0 +1,176 @@
...
+    name   = "name"
+    values = ["runtime-ephemeral-*"]
+  }
+
+  owners = ["${data.aws_caller_identity.current.account_id}"]
+}
+
+locals {
+  suffix = random_id.unique_id.hex
+  tags = tomap({
```
joelmccoy, line 24: would recommend setting these as default tags in the provider like [here](https://github.com/defenseunicorns/uds-aws-account-iac/blob/main/environments/uds-ci-commercial-org/deploy/config.tf#L15-L26)

TristanHoladay: i think i tried this in an earlier iteration and got the 403 error tagging the policy. i agree this is a better way to go though if it works. i'm going to make an issue for this as a follow up PR so we can get this merged and not churn too much as Daniel starts to use it.

UncleGedd (approved):
Great work on this! Deciding to ship and iterate on

PR #158, opened 2024-08-06
fix: empty endpoints are shown as Pending

UncleGedd:
## Description

Previously, endpoints in Pending state were shown as "[Object object]", this updates them to show as "Pending"

Screenshot 2024-08-06 at 3 30 13 PM

### DIFF
--- a/ui/src/lib/components/Link/component.svelte
+++ b/ui/src/lib/components/Link/component.svelte
@@ -7,12 +7,16 @@
   export let text: string
 </script>
 
-<a
-  {href}
-  {target}
-  on:click|stopPropagation
-  rel="noopener noreferrer"
-  class="font-medium text-blue-600 dark:text-blue-500 hover:underline pr-4"
->
-  {text}
-</a>
+{#if href !== ''}
+  <a
+    {href}
+    {target}
+    on:click|stopPropagation
+    rel="noopener noreferrer"
+    class="font-medium text-blue-600 dark:text-blue-500 hover:underline pr-4"
+  >
+    {text}
+  </a>
+{:else}
+  <span class="dark:text-gray-400 mt-4 text-sm">{text}</span>
+{/if}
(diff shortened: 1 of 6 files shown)


ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -250,9 +250,9 @@
                         {/each}
                       </ul>
                     {:else if modifier === 'link-external'}
-                      <Link href={value.href || value} text={value.text || value} target={'_blank'} />
+                      <Link href={value.href || ''} text={value.text || value} target={'_blank'} />
```
TristanHoladay, line 253:
```suggestion
                      <Link href={value.href || ''} text={value.text || ''} target={'_blank'} />
```
might be wrong on this, but i think if we set the href to an empty string instead of value then we're eliminating the possibility to ever send a simple value as the subject of both href and text (e.g. `value = example.com`). so we're enforcing that the data passed to Link must arrive in the shape of `{href: '',text: ''}` if an actual link is to be created.

ui/src/lib/components/Link/component.svelte
```diff
@@ -7,12 +7,16 @@
...
-<a
-  {href}
-  {target}
-  on:click|stopPropagation
-  rel="noopener noreferrer"
-  class="font-medium text-blue-600 dark:text-blue-500 hover:underline pr-4"
->
-  {text}
-</a>
+{#if href !== ''}
```
TristanHoladay, line 10: kind of a nit: i lean toward the guideline that positives are easy to read than negatives. how do you feel about:

```suggestion
{#if href}
```

UncleGedd: easy fix, will implement

ui/src/lib/components/k8s/DataTable/component.svelte
UncleGedd: 👍 discussed sync

PR #159, opened 2024-08-06
chore: refactor logic for ignore fields test helper

UncleGedd:
## Description

Uses lodash to refactor some complicated testing logic. I don't love this approach so I'm keeping it in draft for now


TristanHoladay:
closing cause was implemented in #271

PR #187, opened 2024-08-08
chore(ci): add type checking

UncleGedd:
## Description

add type checking with `npm run check`  to CI and refactors files accordingly


tasks/lint.yaml
```diff
@@ -23,13 +23,11 @@ tasks:
...
     actions:
-      - cmd: npm ci && npm run lint
+      - cmd: npm run lint
         dir: ui
-
-  - name: format-ui
-    description: "Format ui code"
-    actions:
-      - cmd: npm ci && npm run format
+      - cmd: npm run format
```
TristanHoladay, line 30: not sure about putting the format in here since changes won't be committed back to repo in GHA. and if we put it before the lint, which i'm pretty sure will fail on formatting issues, then it will work locally nicely, but i think will give us false passes in CI.

UncleGedd: good point, updated

TristanHoladay (approved):

Issue #180, opened 2024-08-08
Add e2e tests for Table on click events


BillyFigueroa:
### Describe what should be investigated or refactored
Add end to end tests for Table on click events for internal and external links

### Links to any relevant code

PR #188, opened 2024-08-08
feat: adding e2e tests for DataTable

BillyFigueroa:
## Description
Testing the namespace functionality. This can be achieved by clicking on a row which has a namespace action or by selection a namespace from the drop down on the table header

## Related Issue

- #180 

## Screenshots

Here are some screenshots of what the ID's generate to with the code used to generate them. In this case the resource name is being used i.e. Pod to generate ID's specific to this resource on the DataTable. Every table will have unique ID's based on the resource for these 2 items (i.e. header results and drop down option)

```Svelte
<div class="table-header">
  <span class="dark:text-white" data-testid={`${name}-table-header`}>{name}</span>
  {#if isFiltering}
    <span class="dark:text-gray-500 pl-2" data-testid={`${name}-table-header-results`}>
      (showing {$rows.length} of {$numResources})
    </span>
  {:else}
    <span class="dark:text-gray-500 pl-2" data-testid={`${name}-table-header-results`}>({$numResources})</span>
  {/if}
  <div class="relative group">
    <Information class="ml-2 w-4 h-4 text-gray-400" />
    <div class="tooltip tooltip-right min-w-72">
      <div class="whitespace-normal">{description}</div>
    </div>
  </div>
</div>
```
Screenshot 2024-08-08 at 12 49 08 PM

```Svelte
<select id="stream" bind:value={$namespace} data-testid={`${name}-namespace-select`}>
  <option value="" data-testid={`${name}-namespace-select-all`}>All Namespaces</option>
  <hr />
  {#each $namespaces as ns}
    <option value={ns.table.name} data-testid={`${name}-namespace-select-${ns.table.name}`}>
      {ns.table.name}
    </option>
  {/each}
</select>
```

Screenshot 2024-08-08 at 12 48 47 PM

### DIFF
--- a/ui/src/lib/components/k8s/DataTable/component.svelte
+++ b/ui/src/lib/components/k8s/DataTable/component.svelte
@@ -139,13 +139,13 @@
   <div class="table-container">
     <div class="table-content">
       <div class="table-header">
-        <span class="dark:text-white" data-testid={`${name}-table-header`}>{name}</span>
+        <span class="dark:text-white" data-testid="table-header">{name}</span>
         {#if isFiltering}
-          <span class="dark:text-gray-500 pl-2" data-testid={`${name}-table-header-results`}>
+          <span class="dark:text-gray-500 pl-2" data-testid="table-header-results">
             (showing {$rows.length} of {$numResources})
           </span>
         {:else}
-          <span class="dark:text-gray-500 pl-2" data-testid={`${name}-table-header-results`}>({$numResources})</span>
+          <span class="dark:text-gray-500 pl-2" data-testid="table-header-results">({$numResources})</span>
         {/if}
         <div class="relative group">
           <Information class="ml-2 w-4 h-4 text-gray-400" />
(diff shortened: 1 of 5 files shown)


ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -139,11 +139,13 @@
   <div class="table-container">
     <div class="table-content">
       <div class="table-header">
-        <span class="dark:text-white">{name}</span>
+        <span class="dark:text-white" data-testid={`${name}-table-header`}>{name}</span>
```
UncleGedd, line 142: For both `${name}-table-header` and `{name}-table-header-results` below, is it necessary to have a dynamic test id here? There will only be 1 table header and 1 table header results per view. I think the the testid is fine but not sure about the dynamic `${name}` part

UncleGedd: ^  same thing for the rest of the data-testid's in this PR. Thoughts?

UncleGedd (approved):

PR #191, opened 2024-08-09
chore:  go unit tests

decleaver:
## Description

unit tests for compress.go and cache.go


pkg/api/udsmiddleware/compress_test.go
```diff
@@ -0,0 +1,113 @@
...
+			resp := rr.Result()
+			defer resp.Body.Close()
+
+			if encoding := resp.Header.Get("Content-Encoding"); encoding != tt.expectedEncoding {
+				t.Errorf("Expected Content-Encoding %q, got %q", tt.expectedEncoding, encoding)
+			}
+
+			var body []byte
+			var err error
+			if tt.expectedEncoding == "gzip" {
```
UncleGedd, line 58: Instead of using conditionals here, can we use `require` from `"github.com/stretchr/testify/require"`?

pkg/api/udsmiddleware/compress_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if n != len(testContent) {
+		t.Errorf("Expected to write %d bytes, but wrote %d", len(testContent), n)
+	}
+
+	grw.Flush()
+	gzw.Close()
+
+	reader, err := gzip.NewReader(w.Body)
+	if err != nil {
```
UncleGedd, line 100: same comment as above regarding `require` from testify

UncleGedd (approved):
🚀

Issue #170, opened 2024-08-07
Workloads → Pods: In Pods table "Node" and "Workloads" did not populate (shows “-”) 


DannyDTenacious:
Workloads → Pods: In Pods table "Node" and "Workloads" did not populate (shows “-”) 

### Visual Proof (screenshots, videos, text, etc)
<img width="772" alt="Screenshot 2024-08-07 at 2 48 10 PM" src="https://github.com/user-attachments/assets/afc4ab1b-97fc-408b-90d7-88c0d274f220">

TristanHoladay:
@DannyDTenacious can you clarify what "Workloads" are not loading? i didn't see that as a column. I checked `Restarts` and that is showing the correct number when pods have restarts (most don't though). The node is not populating because we're now getting sparse data (no `.spec`) so there's no node name to grab. if i change the call to dense the node name shows, but we wanted to avoid getting the dense data for pods because of the load. i think @jeff-mccoy 's json path PR will allow us to get that field without getting the rest of the dense data, but not sure when that will be ready. should we remove the node column for now?

![Screenshot from 2024-08-07 19-58-02](https://github.com/user-attachments/assets/1745bf3c-c454-43fc-9347-2107a54bc8d0)

DannyDTenacious:
Sorry, that was a mis-type. The only columns that were in questions were "Restart" and "Node" columns. I don't think hiding these columns is needed for now - it shouldn't negatively impact user feedback at this stage. I do think it would be helpful to distinguish the values would be helpful in the short term. Something like "If value = "0" then show "0". If value is not being pulled but will be in the future then show "TBD" or something. Maybe this merits a discussion for the team?

TristanHoladay:
after talking with the others, we ended up putting the nodename as a conditional field in the sparse data so it will show for pods.

PR #194, opened 2024-08-09
fix(ui): restart and node columns in pod table

TristanHoladay:
## Related Issue

- #170 

Screenshot from 2024-08-09 09-45-46

### DIFF
--- a/ui/src/lib/components/k8s/DataTable/component.svelte
+++ b/ui/src/lib/components/k8s/DataTable/component.svelte
@@ -265,7 +265,7 @@
                         {value}
                       </button>
                     {:else}
-                      {value.text || value || (value === 0 && '0') || '-'}
+                      {value.text || (value === 0 ? '0' : value) || '-'}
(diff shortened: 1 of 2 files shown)


ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -265,7 +265,7 @@
                         {value}
                       </button>
                     {:else}
-                      {value.text || value || '-'}
+                      {value.text || value || (value === 0 && '0') || '-'}
```
BillyFigueroa, line 268: What other values besides text and number can we have here?

Would 

```JavaScript
{value.text || (value === 0 ? '0' : value) || '-'}
```

TristanHoladay: ahhh yeah good point. that's probably it. i'll update.

BillyFigueroa (approved):

Issue #156, opened 2024-08-06
Fix Resource Usage Over Time hamburger menu pop-up


BillyFigueroa:
Pop up menu for `Resource Usage Over Time `graph is unreadable

### Steps to reproduce
1. Go to `/overview` view
2. Hover over the hamburger menu item on the `Resource Usage Over Time` graph

### Expected result
We should be able to see the options for the menu

### Actual Result
The container has a background color of white

### Visual Proof (screenshots, videos, text, etc)
<img width="1310" alt="Screenshot 2024-08-06 at 2 12 08 PM" src="https://github.com/user-attachments/assets/375a1e0e-3c77-4067-a7cb-fe05b503ba9e">

PR #195, opened 2024-08-09
fix(ui): fixing issue with graph menu color

BillyFigueroa:
## Description
Fixing issue with graph menu not displaying properly due to background color being white

## Related Issue

- #156 

### Before
Screenshot 2024-08-09 at 12 50 57 PM

### After
Screenshot 2024-08-09 at 12 50 30 PM


ui/src/lib/features/k8s/cluster-overview/styles.postcss
```diff
@@ -0,0 +1,3 @@
+.apexcharts-menu {
+  background-color: rgb(31, 41, 55) !important;
```
UncleGedd, line 2: I'm assuming there was noway to get around the `!important` directive?

BillyFigueroa: Yea, it will not work with out that. I had to override it

UncleGedd (approved):

Issue #77, opened 2024-07-25
API Testing Exploration


adam-defenseunicorns:
### Is your feature request related to a problem? Please describe.
Our current testing process for UDS Runtime lacks comprehensive API testing. This makes it difficult to ensure the reliability and correctness of our API endpoints.

### Describe the solution you'd like
We need to implement automated API testing for UDS Runtime.

**Given** a running Kubernetes cluster with UDS runtime deployed
**When** API endpoints are called with various test cases
**Then** the responses should be validated for correctness, performance, and error handling

Specific requirements:
1. Support for testing RESTful API endpoints
2. Ability to run tests against different environments (dev, staging, production)
3. Integration with CI/CD pipelines
4. Comprehensive test coverage for API endpoints
5. Performance testing capabilities
6. Easy-to-read test reports / failure logs

### Describe alternatives you've considered
1. Manual API testing: This is time-consuming and prone to human error.
2. Using standalone API testing tools: While effective, they may not integrate well with our environments.
3. Writing custom test scripts: This could work but might require significant development effort and maintenance.

UncleGedd:
As part of the API tests, let's also check for duplicated and/or intermediary events (ie. what the debouncing should have fixed)

As an example, when a Pod is deleted, it passes through a few states before eventually terminating, no need to capture all the intermediary states

decleaver:
if having to pick endpoints to test, I think i would go with:
- /api/v1/monitor/pepr
- /api/v1/monitor/pepr/{stream}
- /api/v1/monitor/cluster-overview

(any corev1 type endpoints i.e. pods)
- /api/v1/resources/workloads/pods (dense/sparse, once=True/False)
- /api/v1/resources/workloads/pods/{uid} (dense/sparse)

(any custom resource endpoint i.e. UDSPackage)
- /api/v1/resources/configs/uds-packages (dense/sparse, once=True/False)
- /api/v1/resources/configs/uds-packages/{uid} (dense/sparse)

I think this would cover the different handler types and resource types

PR #198, opened 2024-08-14
chore: api testing adr

UncleGedd:
## Description

adds an API testing ADR

## Related Issue

- #77

### DIFF
--- a/adr/0001-api-testing.md
+++ b/adr/0001-api-testing.md
@@ -1,0 +1,46 @@
+# 1. API Testing
+
+Date: 14 August 2024
+
+## Status
+
+In-progress
+
+## Context
+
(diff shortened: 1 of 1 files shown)


decleaver (changes requested):
I think also worth testing:
/api/v1/monitor/pepr
/api/v1/monitor/pepr/{stream}
/api/v1/monitor/cluster-overview

since they are handled differently than the mentioned bind groupings

UncleGedd:
> I think also worth testing: /api/v1/monitor/pepr /api/v1/monitor/pepr/{stream} /api/v1/monitor/cluster-overview
> 
> since they are handled differently than the mentioned bind groupings

updated!

decleaver (changes requested):
thoughts on breaking it down even more and just testing a corev1 type and a custom resource type? I think the handlers for the corev1 types are all the same, and same thing with the custom resources. Would also cover the dynamic and shared informer paths

decleaver (approved):
👍

PR #202, opened 2024-08-15
chore: add inputs to actions setup

TristanHoladay:
## Description
add inputs to actions setup so that we can customize for each workflow which tools we want to install. this should speed up CI a little bit in some places.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -1,15 +1,44 @@
 name: Setup Environment
 description: Environment Setup
 
+inputs:
+  go:
+    description: 'Go version'
+    required: false
+    default: "true"
+  node:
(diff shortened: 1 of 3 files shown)


UncleGedd:
Looks like the performance gain here is from not always running the docker buildx and QEMU installs. Those actually install quickly so the gain is minimal. Personally, I find workflows hard to reason through anyway, so unless the extra logic can remove an appreciable amount of time (maybe >10s?) I'd keep it simple and not add the extra logic

TristanHoladay:
@UncleGedd that's a good point. i think the only place this really saves time atm is the nightly-infra workflow, which doesn't really matter at that point. so yeah i'm cool with closing this and implementing in the future if it makes sense for more workflows.

Issue #186, opened 2024-08-08
Discuss replacing svelte-apexcharts dependency


UncleGedd:
### Describe what should be investigated or refactored

The [svelte-apexcharts](https://github.com/galkatz373/svelte-apexcharts) dep hasn't been updated in 4 years. Discuss with @jeff-mccoy if we swap this for something else

BillyFigueroa:
# Options
Looked into some options. I did not do all the research likely needed but I did try to find the best options I could find. Some of more popular and known and some are smaller but more actively maintained

## Here is the list

- [ ] https://github.com/mhkeller/layercake
- [ ] https://github.com/SauravKanchan/svelte-chartjs (based off https://www.chartjs.org/)
- [ ] https://www.layerchart.com (https://www.layerchart.com/docs/examples/Area#multiple_series_(using_overrides))
- [ ] https://github.com/apache/echarts (https://echarts.apache.org/en/index.html)

There is also this…
- [ ] https://www.sveltecharts.com/

but this seems like individual graphs which could present it’s own problem for maintenance

We can also just try to find a typescript library and add any potential reactivity that needs to be added. Something like this?
- [ ] https://github.com/apache/echarts

PR #206, opened 2024-08-15
feat: adding echarts

BillyFigueroa:
## Description
Testing `Echarts` to replace `svelte-apexcharts`

### About Echartjs
Echarts JS is a powerful charting and vizualization library which uses a lightway canvas library called [zrender](https://github.com/ecomfe/zrender). It has 59.9k starts on github

This library is being used directly, rather than using a svelte wrapper because the svelte wrapper option is too small (1 real contributor) which might lead to an issue with maintenance in the future

#### Pros
- Powerful
- Uses a lightweight Canvas library
- Easy to make responsive
- Easy to customize

#### Cons
- With so many options comes a bigger file size

## Related Issue
- #186 

## Screenshots

Screenshot 2024-08-15 at 1 43 06 PM


jeff-mccoy (commented):
I dont think we'll be able to consider echarts based on the contributor history

BillyFigueroa:
> I dont think we'll be able to consider echarts based on the contributor history

ah that sucks! but makes total sense. This one was much nicer to work with. I ll go ahead and close this one out

PR #212, opened 2024-08-16
feat: make root URL the overview page

UncleGedd:
## Description

Adding redirect from `/` to `/overview`


BillyFigueroa:
@UncleGedd a test seems to be failing. Fix that up so I can approve

Also, why are we redirecting instead of just moving the code in the overview page to the home page?

UncleGedd:
> @UncleGedd a test seems to be failing. Fix that up so I can approve
> 
> Also, why are we redirecting instead of just moving the code in the overview page to the home page?

I could go either way, the redirect was to have the URL show where you are in the app. Since 'Overview' is on the sidebar I thought it made sense to add it to the URL for uniformity

Just chatted with @decleaver and @TristanHoladay and we agreed to move it to the root

TristanHoladay (approved):

Issue #222, opened 2024-08-19
runtime-canary: better debugging and resiliency


UncleGedd:
### Is your feature request related to a problem? Please describe

- ensure we can log in using SSM through the AWS console to debug k3d issues in the runtime-canary env
- when watching the rollout of UDS Core + Runtime, we saw the promtail "too many files" error, may need to adjust the `ulimit` but be careful because it may only persist per session
- add docs for deploying locally

PR #223, opened 2024-08-19
chore: update test iac

### LINKED ISSUE #211: Update tagging in nightly ephemeral IAC
@joelmccoy commented in #157 that we should move local tags to default tags in the provider for our nightly ephemeral cluster

 https://github.com/defenseunicorns/uds-runtime/pull/157#discussion_r1720072864

TristanHoladay:
## Description
Update tagging of IAC, add SSM for debugging, and set ulimits in AMI to allow for Core to deploy without issue.

## Related Issue

Resolves #211 #222

### DIFF
--- a/.github/test-infra/README.md
+++ b/.github/test-infra/README.md
@@ -1,0 +1,66 @@
+# Runtime Ephemeral Infrastructure
+
+The UDS Runtime IAC is used by the [nightly-infra workflow](../workflows/nightly-infra.yaml), via [uds tasks](./tasks/infra.yaml), to destroy and create ephemeral testing clusters, using the latest `nightly-unstable` image of UDS Runtime.
+
+## How it Works
+
+When the nightly workflow kicks off, it will `init` using the backend variables defined in the workflow to then destroy the currently running ec2 instance and cluster. After removing the old instance, it will create a new ec2 instance in the UDS CI AWS account, that on startup will do the following:
+
+1. clone the [uds-k3d](https://github.com/defenseunicorns/uds-k3d) repo, setting `nginx.conf` to redirect for the `.burning.boats` domain
+1. run the default task of `uds-k3d`, creating the k3d cluster on the instance
+1. setup the `kubecontext` to be used by `uds`
+1. pull the `.burning.boats` tls cert and key from secrets manager
+1. deploy the `init` and `UDS Core` packages
+1. deploy the `UDS Runtime` package
+
+## Custom AMI
+
+The ec2 instance is created with a custom AMI. We use `packer` to define the AMI in [runtime.pkr.hcl](./packer/runtime.pkr.hcl) and build / push it to our AWS accounts.
+
+## Development and Testing
+
+> **NOTE**  
+> **Please use the UDS Dev AWS Account instead of CI**
+
+For local development and testing:
+
+pre-requisites:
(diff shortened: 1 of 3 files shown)


.github/test-infra/README.md
```diff
@@ -0,0 +1,66 @@
+# Runtime Ephemeral Infrastructure
```
decleaver, line 1: appreciate the docs! thoughts on moving it into just the docs folder vs in here?

UncleGedd: I can def see why moving it to `docs` makes sense today, but in the future the `docs` folder will likely become the markdown for the docs website. These docs are more dev-focused as opposed to "public-facing" so IMO they make sense closer to the code they're documenting

.github/test-infra/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The UDS Runtime IAC is used by the [nightly-infra workflow](../workflows/nightly-infra.yaml), via [uds tasks](./tasks/infra.yaml), to destroy and create ephemeral testing clusters, using the latest `nightly-unstable` image of UDS Runtime.
+
+## How it Works
+
+When the nightly workflow kicks off, it will `init` using the backend variables defined in the workflow to then destroy the currently running ec2 instance and cluster. After removing the old instance, it will create a new ec2 instance in the UDS CI AWS account, that on startup will do the following:
```
UncleGedd, line 7:
```suggestion
When the nightly workflow kicks off, it will `tofu init` using the backend variables defined in the workflow, then destroy the currently running EC2 instance and related infra. After removing the old infra, it will create a new EC2 instance in the UDS CI AWS account, that on startup will do the following:
```

.github/test-infra/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. clone the [uds-k3d](https://github.com/defenseunicorns/uds-k3d) repo, setting `nginx.conf` to redirect for the `.burning.boats` domain
+1. run the default task of `uds-k3d`, creating the k3d cluster on the instance
+1. setup the `kubecontext` to be used by `uds`
+1. pull the `.burning.boats` tls cert and key from secrets manager
+1. deploy the `init` and `UDS Core` packages
+1. deploy the `UDS Runtime` package
+
+## Custom AMI
+
+The ec2 instance is created with a custom AMI. We use `packer` to define the AMI in [runtime.pkr.hcl](./packer/runtime.pkr.hcl) and build / push it to our AWS accounts.
```
UncleGedd, line 18: Can you specifically mention the command used to build/push? I think it's just `packer build runtime.pkr.hcl` (this keeps me from grepping my cmd history)

.github/test-infra/README.md
TristanHoladay: i agree with @UncleGedd but I think what you're bringing up @decleaver is a legit ask because it will increase visibility / findability. Maybe a good way to handle this is to mention these docs in the `CONTRIBUTING.md` so that devs can have an easier time finding them.

decleaver: sounds good, yeah my big thing was just making it easy to find/know it exists. You are able to create a subfolder under docs and that doesn't get pulled into the website if we decide in the future to go that route.

UncleGedd (approved):

PR #225, opened 2024-08-19
chore: update readme with badges

TristanHoladay:
## Description
Add badges to README and update workflow titles to show up as more readable.


README.md
```diff
@@ -1,6 +1,15 @@
-# UDS Runtime
+# UDS Runtime - providing views and insights into your UDS cluster
```
UncleGedd, line 1: IMO we should keep the description on the line below and leave the header as just "UDS Runtime". I don't see many GH projects that have the description in the header (ie. [kubewall](https://github.com/kubewall/kubewall), [pepr](https://github.com/defenseunicorns/pepr)). If we want something in the header, let's make it short and sweet like [Zarf](https://github.com/zarf-dev/zarf) or [Lula](https://github.com/defenseunicorns/lula))

UncleGedd (approved):

PR #231, opened 2024-08-20
chore(api): add api tests

### LINKED ISSUE #77: API Testing Exploration

TristanHoladay:
## Description
Adds api tests according the the ADR. 

Inspired by the work started in #166 by @UncleGedd 

## Related Issue

Resolves #77


pkg/api/start.go
```diff
@@ -160,22 +159,18 @@ func Start(assets embed.FS) error {
...
-	staticFS, err := fs.Sub(assets, "ui/build")
-	if err != nil {
-		return fmt.Errorf("failed to create static file system: %w", err)
-	}
-
-	if err := fileServer(r, http.FS(staticFS)); err != nil {
-		return fmt.Errorf("failed to serve static files: %w", err)
-	}
+	if assets != nil {
+		staticFS, err := fs.Sub(assets, "ui/build")
```
decleaver, line 163: good call on the extra nil checks. (Nitpick) make this call a one liner like how you are handling fileServer below.

pkg/api/start.go
```diff
@@ -160,22 +159,18 @@ func Start(assets embed.FS) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create static file system: %w", err)
+		}
 
-	log.Println("Starting server on :8080")
-	//nolint:gosec
-	if err := http.ListenAndServe(":8080", r); err != nil {
```
decleaver, line 174: curious on why you moved this to main

pkg/test/api_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	tests := []struct {
+		name    string
+		url     string
+		isDense bool
+	}{
+		{
+			name: "once=true",
+			url:  "/api/v1/resources/workloads/pods?once=true",
+		},
```
decleaver, line 31: add case for once=true and dense=true, you are already handling sse sparse and dense

pkg/api/start.go
TristanHoladay: cause @UncleGedd did lol. i believe it's because actually calling `ListenAndServe` interferes with the tests. i know when i worked on this before seeing Gedd's spike branch i had problems with it. I could go back though and experiment again and see if it's actually necessary to move out.

pkg/test/api_test.go
TristanHoladay: just to clarify you're saying to add a case for `/pods?once=true&dense=true` right?

pkg/api/start.go
TristanHoladay: i like that idea, but looks like it won't work cause staticFS is used below in the fileServer call.

pkg/api/start.go
TristanHoladay: yeah just tried running the `ListenAndServe` from within `Setup()` and the test kick off, starts serving on port 8080, and then hangs until it errors out. Something about goroutines waiting indefinitely. I tried the setup call outside and inside the goroutine declared in the test and it failed both ways. maybe the issue is that the server is now running and listening in the foreground, blocking the rest of the test from running.

TristanHoladay:
with the use of the build tags for separating tests, the vscode go extension test explorer fails to run or debug the tests. I update the extension's settings.json with `"go.testTags": "unit,integration",` and test explorer started working again.

pkg/test/api_test.go
```diff
@@ -0,0 +1,248 @@
+//go:build integration
```
decleaver, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// SPDX-FileCopyrightText: 2024-Present The UDS Authors

//go:build integration
```

decleaver:
Issue for follow on field selection tests https://github.com/defenseunicorns/uds-runtime/issues/252

decleaver (approved):
lgtm, added issue for follow on work

Issue #233, opened 2024-08-21
Add go binaries as release artifacts


decleaver:
Add the go binaries for the different os/architectures similarly to what zarf and uds-cli are currently doing.

These binaries will end up getting used by uds-cli for the `uds ui` command to run UDS runtime locally.

PR #240, opened 2024-08-21
chore:  add go binaries as artifacts of releases

decleaver:
## Description

add go binaries as artifacts of releases

## Related Issue

- #233


tasks/setup.yaml
```diff
@@ -8,15 +8,34 @@ tasks:
       - cmd: go mod tidy
       - cmd: CGO_ENABLED=0 go build -o ./build/uds-runtime .
 
+  - name: build-all
```
TristanHoladay, line 11: I think we'll need the `build-ui` task here, like in `build:push-container` since this task is called before that one in the workflows above. Might be worth looking at too if that task should be refactored if we're doing all the builds in this one.

decleaver: good catch with the UI! Don't think we can use build-all in the push-container task since that task only wants the linux builds, doesn't care about darwin.

TristanHoladay (dismissed):

release-please-config.json
```diff
@@ -14,7 +14,8 @@
         "zarf.yaml",
         "tasks.yaml",
         "tasks/build.yaml",
-        "chart/values.yaml"
+        "chart/values.yaml",
+        "build/**"
```
UncleGedd, line 18: Does `extra-files` allow glob patterns? Also what in `build` are we updating with release please?

decleaver: Not immediately apparent from the release-please documentation, but the machines tell me it is valid. Putting all of the binaries inside of the build dir

UncleGedd (approved):

Issue #235, opened 2024-08-21
API auth


decleaver:
Add API auth for when running UDS Runtime locally

Context from Jeff:
get uds ui up and running (an be feature-flagged), but letting users play with that would be great

--caution: need to consider auth for that, if we're on localhost binding like we have rn thats fine, but we'll want to do some type of dynamic api key to prevent random other processes from accessing via the same port, see https://github.com/defenseunicorns/zarf-ui/blob/main/src/api/auth/token.go and https://github.com/defenseunicorns/zarf-ui/blob/main/src/ui/routes/auth/%2Bpage.svelte for examples that might help

NOTE: This issue is covers the API auth piece, the uds ui piece is covered by an issue in the uds-cli repo

PR #245, opened 2024-08-22
feat: local mode token protection

decleaver:
## Description

Implements api authentication.

The api generates an API token which is then used by the UI to authenticate. This will be set by default, with the option to disable (for in-cluster deployments)

## Related Issue

- #235


docs/api-auth.md
```diff
@@ -0,0 +1,13 @@
+API AUTHENTICATION
```
UncleGedd, line 1:
```suggestion
# API AUTHENTICATION
```

go.mod
```diff
@@ -15,6 +15,8 @@ require (
 	k8s.io/metrics v0.30.3
 )
 
+require github.com/defenseunicorns/pkg/helpers v1.1.1 // indirect
```
UncleGedd, line 18: can we move this into one of the other `require` blocks?

docs/api-auth.md
```diff
@@ -0,0 +1,13 @@
...
+
+API authentication has been implemented for when running locally, to prevent unauthorized access to the API from other processes. The API uses a token-based authentication system. The token is generated by the backend server and is used to authenticate the user. API authentication is enabled based on the value of the `VITE_API_AUTH` environment variable.
```
UncleGedd, line 3:
```suggestion
API authentication is used prevent unauthorized access to the API from other processes when running UDS Runtime locally. The API uses a token-based authentication system. The token is generated by the backend server and is used to authenticate the user. API authentication is enabled based on the value of the `VITE_API_AUTH` environment variable.
```

pkg/api/start.go
```diff
@@ -30,6 +35,22 @@ import (
...
+	port := "8080"
+
+	ip := "127.0.0.1"
+
+	// If the env variable API_TOKEN is set, use that for the API secret
+	token := os.Getenv("API_TOKEN")
+	var err error
+	// Otherwise, generate a random secret
+	if token == "" {
+		token, err = auth.RandomString(96)
```
UncleGedd, line 48: Any reason for 96 in particular?

pkg/api/start.go
```diff
@@ -30,6 +35,22 @@ import (
 // @BasePath /api/v1
 // @schemes http https
 func Start(assets embed.FS) error {
+	apiAuth := os.Getenv("VITE_API_AUTH")
```
UncleGedd, line 38: I think I'm getting hung up on the mention of `VITE`. The backend doesn't know anything about our frontend build framework.

pkg/api/start.go
```diff
@@ -44,7 +65,14 @@ func Start(assets embed.FS) error {
 
 	// Add Swagger UI route
 	r.Get("/swagger/*", httpSwagger.WrapHandler)
+	// expose env vars to frontend via endpoint
+	r.Get("/config", serveConfig)
```
UncleGedd, line 69: Is this for dev'ing only? I imagine when running from the `uds ui` command this wouldn't be necessary (or when running remotely)

go.mod
decleaver: weird... go mod tidy put it there, moved it down below.

pkg/api/start.go
decleaver: nope, just a big number (high entropy and security). Was copied over from the work previously done on zarf ui.

pkg/api/start.go
decleaver: renaming to API_AUTH_ENABLED

pkg/api/start.go
decleaver: this endpoint will get used in all cases to determine at runtime if API_AUTH_ENABLED. Will update to make more clear.

docs/api-auth.md
```diff
@@ -0,0 +1,13 @@
+# API AUTHENTICATION
+
+API authentication is used prevent unauthorized access to the API from other processes when running UDS Runtime locally. The API uses a token-based authentication system. The token is generated by the backend server and is used to authenticate the user. API authentication is enabled based on the value of the `API_AUTH_ENABLED` environment variable.
```
TristanHoladay, line 3:
```suggestion
API authentication is used to prevent unauthorized access to the API from other processes when running UDS Runtime locally. The API uses a token-based authentication system. The token is generated by the backend server and is used to authenticate the user. API authentication is enabled based on the value of the `API_AUTH_ENABLED` environment variable.
```

pkg/api/start.go
```diff
@@ -44,7 +65,16 @@ func Start(assets embed.FS) error {
...
 	r.Get("/swagger/*", httpSwagger.WrapHandler)
+	// expose API_AUTH_ENABLED env var to frontend via endpoint
+	r.Get("/config", serveConfig)
 	r.Route("/api/v1", func(r chi.Router) {
+		// Require a valid token for API calls
+		if strings.EqualFold(apiAuth, "true") {
+			// If api auth is enabled, require a valid token for all routes under /api/v1
+			r.Use(auth.RequireSecret(token))
+			// Endpoint to test if connected with auth
+			r.Head("/", auth.Connect)
```
TristanHoladay, line 76: i think @UncleGedd brought this us up earlier, but do we need this connect test?

ui/src/lib/features/k8s/events/store.ts
```diff
@@ -19,7 +21,12 @@ export type Columns = ColumnWrapper<Row>
 
 export function createStore(): ResourceStoreInterface<Resource, Row> {
   // Using dense=true because most of the fields are stripped out in the default spareResource stream
-  const url = `/api/v1/resources/events?dense=true`
+  let url = `/api/v1/resources/events?dense=true`
+
+  if (get(apiAuthEnabled)) {
```
TristanHoladay, line 26: is this needed now that the store start logic adds an `&` if there's already a `?` ?

pkg/api/start.go
decleaver: nope, if you guys like I can leverage an existing endpoint. I have it as a separate endpoint just so that the api auth logic is separate from any of the other resources. For example, Its possible that the pods or nodes endpoints can be broken but api auth is still good.

ui/src/lib/features/k8s/store.ts
```diff
@@ -167,13 +167,20 @@ export class ResourceStore<T extends KubernetesObject, U extends CommonRow> impl
...
-    this.#eventSource = new EventSource(this.url)
+
+    if (!get(apiAuthEnabled)) {
+      this.#eventSource = new EventSource(this.url)
+    } else {
+      const apiToken: string = sessionStorage.getItem('token') ?? ''
+      // Check if the URL already contains a '?' for urls with multiple search params
+      const separator = this.url.includes('?') ? '&' : '?'
+      this.#eventSource = new EventSource(`${this.url}${separator}token=${apiToken}`)
+    }
```
TristanHoladay, line 183: what do you think about something like this to get rid of the negative assertion and simplify a bit?

```suggestion
   if (get(apiAuthEnabled)) {
      const apiToken: string = sessionStorage.getItem('token') ?? ''
      // Check if the URL already contains a '?' for urls with multiple search params
      const separator = this.url.includes('?') ? '&' : '?'
      this.url = `${this.url}${separator}token=${apiToken}`
    }
    
    this.#eventSource = new EventSource(this.url)
```

ui/tests/api-auth.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await page.getByRole('link', { name: 'Overview' }).click()
+    const nodeCountEl = page.getByTestId(`node-count`)
+    await expect(nodeCountEl).toHaveText('1')
+  })
+
+  test('pod view access', async ({ page }) => {
+    await page.goto(`/auth?token=${extractedToken}`)
+    await page.getByRole('button', { name: 'Workloads' }).click()
+    await page.getByRole('link', { name: 'Pods' }).click()
+    const element = page.locator(`.emphasize:has-text("podinfo")`)
```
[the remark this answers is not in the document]
decleaver: this one might be tricky, since the emphasize class is being applied to all of the table data elements. I think this is why this approach was taken in some other tests. Also possible I just dont know, so more than happy to update if you have some ideas here.

ui/src/lib/features/k8s/events/store.ts
decleaver: Good catch!!

ui/src/lib/features/navigation/navbar/component.svelte
```diff
@@ -4,6 +4,17 @@
 <script lang="ts">
   import { NotificationFilled } from 'carbon-icons-svelte'
   import { isSidebarExpanded } from '../store'
+  import { authenticated } from '$lib/features/api-auth/store'
+  import { apiAuthEnabled } from '$lib/features/api-auth/store'
+
+  // Don't expand sidebar if on api auth page
+  $: {
```
TristanHoladay, line 11: not sure you need this given the conditional below. i removed it and i can't see the side nav on the auth page. it does flash on reload if you're looking for it, but it flashes even with this reactive conditional check.

decleaver: yeah, so that covers the (non-visible) sidebar being expanded when you have a bad api key. You don't see the sidebar either way, but if you remove the `${` the text and doug image get pushed to the right

TristanHoladay: ohhh. coolio.

ui/src/lib/features/k8s/events/store.ts
```diff
@@ -19,7 +19,7 @@ export type Columns = ColumnWrapper<Row>
 
 export function createStore(): ResourceStoreInterface<Resource, Row> {
   // Using dense=true because most of the fields are stripped out in the default spareResource stream
-  const url = `/api/v1/resources/events?dense=true`
+  let url = `/api/v1/resources/events?dense=true`
```
TristanHoladay, line 22: nit: can be a const again

ui/src/lib/utils/http.ts
```diff
@@ -0,0 +1,68 @@
...
+      this.invalidateAuth()
+    }
+  }
+
+  // Updates the internal token used for authentication.
+  updateToken(token: string) {
+    sessionStorage.setItem('token', token)
+  }
+
+  private invalidateAuth() {
```
TristanHoladay, line 27: i think you can use the `#` syntax here, like in the main store.ts.

```suggestion
  #invalidateAuth() {
```

ui/src/routes/+page.svelte
```diff
@@ -3,6 +3,22 @@
...
+  import { updateApiAuthEnabled } from '$lib/utils/http'
+  import { onMount } from 'svelte'
+
+  onMount(async () => {
+    updateApiAuthEnabled()
+  })
+
+  // Redirect to /auth if api auth is enabled and user is not authenticated
+  $: if ($apiAuthEnabled && !$authenticated) {
+    window.location.href = '/auth'
```
TristanHoladay, line 17:
```suggestion
        goto('/auth')
```

ui/src/lib/features/k8s/events/store.ts
decleaver: lol yeah, linter didnt like it either

ui/src/lib/utils/http.ts
decleaver: whoops, didnt like that one. Need it to be private, otherwise not able to call it in the constructor above.

TristanHoladay: sorry if you already tried this, but in the constructor will `this.#invalidateAuth()` work?

ui/src/routes/auth/+page.svelte
```diff
@@ -0,0 +1,46 @@
...
+  import { authenticated } from '$lib/features/api-auth/store'
+  import { apiAuthEnabled } from '$lib/features/api-auth/store'
+  import { updateApiAuthEnabled } from '$lib/utils/http'
+
+  export let data
+
+  let authFailure = false
+
+  onMount(async () => {
+    authenticated.set(false)
```
TristanHoladay, line 17: is this needed since the store defaults to start as false and you're setting it to false if it fails below as well?

ui/src/lib/utils/http.ts
decleaver: ah nice, yeah it liked that

ui/src/routes/auth/+page.svelte
decleaver: removed

ui/tests/api-auth.spec.ts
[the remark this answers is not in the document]
BillyFigueroa: @decleaver I can help with this if you want. I can change the table column to have an id so it can be targeted. I would also update the other tests. Do you want me to do this in a different PR or work with you on your PR?

pkg/api/start.go
TristanHoladay: i understand now how you were using this. lgtm

ui/src/lib/features/k8s/store.ts
```diff
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2024-Present The UDS Authors
 
+import { apiAuthEnabled } from '$lib/features/api-auth/store'
```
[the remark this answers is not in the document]
BillyFigueroa: I ll take a look at what the options might be for that

ui/tests/api-auth.spec.ts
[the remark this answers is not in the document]
decleaver: lol this pr is blowing up, if you want to do it on a separate branch, then I can use the id in this test case

TristanHoladay (approved):

PR #246, opened 2024-08-23
chore: swap svelte-chartjs for chartjs lib

BillyFigueroa:
## Description
Testing `chartjs` in vanilla JS to replace `svelte-apexcharts`

### About chartjs
Chart JS is a lightweight charting library which uses canvas to generate the graphics. It has 64k+ starts on github

In combination with Chart.js we are also using svelte-chartjs which is a wrapper around chartjs for svelte

#### Pros
- Lightweight
- Uses Canvas
- Easy to make respnsive

#### Cons
- Customizing specific aspects of the graph is not so simple. It takes understanding or being familiar with the API
- Had to use a svelte library to make for a simpler integration
- Because it's lightweight, sometimes plugins have to be used to enhance it

## Related Issue
- #186 

#### Side by Side
Screenshot 2024-08-23 at 1 37 28 PM

#### Before
Screenshot 2024-09-30 at 11 58 28 AM

#### After (diff data sets)
Screenshot 2024-09-30 at 11 58 36 AM


ui/src/lib/features/k8s/cluster-overview/component.svelte
```diff
@@ -50,155 +50,133 @@
...
-    apexChart?.updateOptions(options)
-  }
-
-  function updateClusterData(clusterData: ClusterData): ApexOptions {
-    return {
-      ...options,
-      series: [
-        {
-          name: 'CPU Usage',
-          data: (clusterData.historicalUsage ?? []).map((point) => ({
```
UncleGedd, line 185: Need to be careful here that `?? []` fixed a gnarly bug where the chart wasn't loading

BillyFigueroa: > Need to be careful here that, `?? []` fixed a gnarly bug where the chart wasn't loading

This code was removed for a completely different library. The diff here couldn't understand it was different code and it looks like the code was updated. It was taken out. The new code sets labels and data to empty arrays here

Screenshot 2024-09-30 at 6 03 20 PM

And they are updated inside the `onMessage` call here

Screenshot 2024-09-30 at 6 04 47 PM

UncleGedd: Looks good on my end, tested with a large cluster (which was what caused the bug previously)

UncleGedd (approved):

PR #248, opened 2024-08-26
fix(ci): swag init task output directory

TristanHoladay:
## Description
make sure swag init puts swagger docs in /pkg/api/docs


decleaver (approved):
lgtm

PR #249, opened 2024-08-26
fix: playing with field selectors

UncleGedd:
## Description

using client-go `jsonpath` + 🤖 to slim down filter logic


jeff-mccoy:
I probably have my stash for my jsonpath work, the issue was either parsing union types or trying to pick multiple keys without duplicating data

PR #250, opened 2024-08-27
feat(api): strip annotations from sparse resources

jeff-mccoy:
## Description

This PR strips annotations from sparse datasets to avoid excess data and prevent leaky secret data with the `kubectl.kubernetes.io/last-applied-configuration` [annotation](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/). It also corrects the mutation of the original object by `extractSparseObject` by making a deep copy of the object metadata prior to modifying it. Removal of `managedFields` was moved to the source function to affect all resource views.

### DIFF
--- a/pkg/api/resources/list.go
+++ b/pkg/api/resources/list.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


UncleGedd (approved):
🚀

TristanHoladay (approved):

PR #254, opened 2024-08-27
refactor(ui): updating tests to use data-testid's

BillyFigueroa:
## Description
Refactoring playwright tests to use `data-testid`'s instead of `page.locator`


ui/tests/navigation.spec.ts
```diff
@@ -50,16 +44,14 @@ test.describe('Navigation', async () => {
       await page.getByRole('button', { name: 'Workloads' }).click()
       await page.getByRole('link', { name: 'Pods' }).click()
 
-      const element = page.locator(`.emphasize:has-text("${query}")`)
-      await expect(element).toBeVisible()
+      await expect(page.getByTestId('coredns-77ccd57875-htl2l-testid-1')).toHaveText('coredns-77ccd57875-htl2l')
```
decleaver: yeah, you wont be able to use that id (77ccd57875-htl2l) after the first "-" since it will be different every time its deployed

BillyFigueroa: @decleaver so I need to leave that one as is?

ui/tests/navigation.spec.ts
```diff
@@ -50,16 +44,16 @@ test.describe('Navigation', async () => {
       await page.getByRole('button', { name: 'Workloads' }).click()
       await page.getByRole('link', { name: 'Pods' }).click()
 
-      const element = page.locator(`.emphasize:has-text("${query}")`)
-      await expect(element).toBeVisible()
+      await expect(page.getByTestId('admin-ingressgateway-65c568569f-qqxk2-testid-1')).toHaveText(
```
decleaver, line 47: This won't work because the id changes everytime a pod is deployed. The value after the first "-" (65c568569f-qqxk2) will be different each time.

ui/src/lib/features/navigation/sidebar/component.svelte
```diff
@@ -46,6 +46,7 @@
 
 <aside
   id="main-sidebar"
+  data-testid="main-sidebar-test-id"
```
TristanHoladay, line 49: is it possible to remove the `-test-id` from the testid string, or is that needed cause of conflicts with the `id`?

BillyFigueroa: @TristanHoladay we are using the id right above it to style. Would you like a different name or to remove the id and just have `data-testid="main-sidebar"`. I kept it there because it is being used to target css but it can be removed and updated to use somehting like `aside[data-testid="main-sidebar"]`

ui/tests/navigation.spec.ts
BillyFigueroa: > This won't work because the id changes everytime a pod is deployed. The value after the first "-" (65c568569f-qqxk2) will be different each time.

Thanks. I ll try to see if there is a way to find using "starting with admin-ingressgatewat". Is that going to be unique in the pods? just one admin-ingressgateway row?

decleaver: yeah, should just be one

ui/src/lib/features/navigation/sidebar/component.svelte
TristanHoladay: if the id is used for targeting css then i guess we don't want to remove it, but it's a little meh to have `test-id` in the testid value. you know just being redundant. i'm cool with whatever you think is best though.

BillyFigueroa: @TristanHoladay I updated the PR to use only one ID and changes the CSS locator.

ui/tests/navigation.spec.ts
BillyFigueroa: @decleaver I reverted this test. This is one case we will have to use `page.locator` instead of anyone of the defined locators etc

decleaver (approved):
👍

Issue #129, opened 2024-08-01
Status Colors


Madeline-UX:
### Is your feature request related to a problem? Please describe

As a DevOps engineer or system administrator,
I want to see color-coded status indicators for all Kubernetes resources in our application,
So that I can quickly assess the health and state of our infrastructure at a glance.

### Acceptance Criteria

The application uses a consistent color scheme for all Kubernetes resource statuses:

<img width="400" alt="Screenshot 2024-08-01 at 3 30 57 PM" src="https://github.com/user-attachments/assets/907feda2-03a8-4804-b653-4dfd01c99822">

- [ ] Green for success/healthy states
- [ ] Blue for informational/ongoing processes
- [ ] Orange for warnings and pending states
- [ ] Red for errors and critical issues
- [ ] Grey for disabled or inactive resources

1. The color scheme is applied uniformly across all resource types (Pods, Deployments, Services, etc.) as per the provided recommendations.
2. Each resource in the UI displays its status using the appropriate color.
3. Color indicators are accompanied by text labels for clarity and accessibility.

### Design Resources

[Color Status Design Guidelines](https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=3723-2949&t=htuNy6u4cok3uBiu-1)

adam-defenseunicorns:
@jeff-mccoy is this something we can toss back into the backlog for someone else to take on?

BillyFigueroa:
> @jeff-mccoy is this something we can toss back into the backlog for someone else to take on?

Is this ticket front end only?

PR #264, opened 2024-08-29
feat(ui): adding status color mapping for k8s

BillyFigueroa:
## Description
Adding color mapping for the various resources for k8s

## Related Issue

- #129 

## Screenshots
Screenshot 2024-08-29 at 1 32 23 PM


ui/src/lib/features/k8s/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+}
+
+// Function to get the color and status for a specific type and status
+export const getColorAndStatus = <T extends keyof K8StatusMapping>(
+  type: T,
+  status: keyof K8StatusMapping[T],
+): string => {
+  return ((k8StatusMapping[type][status] as { color: string }).color as string) || 'Unknown'
+}
```
TristanHoladay, line 86:
```suggestion
export const getColorAndStatus = <T extends keyof K8StatusMapping>(
  type: T,
  status: keyof K8StatusMapping[T],
) => {
  return ((k8StatusMapping[type][status] as { color: string }).color) || 'Unknown'
}
```

ui/src/lib/features/k8s/workloads/pods/status/component.test.ts
```diff
@@ -6,11 +6,11 @@ describe('Status component', () => {
...
     expect(container.firstChild).toHaveClass('text-green-400')
   })
-  test('renders text-red-400 for Running', () => {
-    const { container } = render(Status, { props: { status: 'Stopped' } })
+  test('renders text-red-400 for Failed', () => {
+    const { container } = render(Status, { props: { status: 'Failed' } })
     expect(container.firstChild).toHaveClass('text-red-400')
   })
-  test('renders text-yellow-500 for Running', () => {
+  test('renders text-yellow-500 for Pending', () => {
```
TristanHoladay, line 13:
```suggestion
  test('renders text-orange-300 for Pending', () => {
```

ui/src/lib/features/k8s/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type StatusColorMapping = {
+  success: string
+  info: string
+  warning: string
+  error: string
+  disabled: string
+}
+
+const statusColors: StatusColorMapping = {
```
TristanHoladay, line 11: what do you think about getting rid of the `StatusColorMapping` type and instead doing `as const` on `statusColors` so we can see the class string in intellisense?

Screenshot from 2024-08-30 08-40-13

Screenshot from 2024-08-30 08-38-11

ui/src/lib/features/k8s/helpers.ts
BillyFigueroa: Nice little cleanup. I ll make the change

ui/src/lib/features/k8s/helpers.ts
BillyFigueroa: I just watched a video this morning on how you can use `as const` in this way to sort of replicate enum's as a way to not use enums

I ll make the change

TristanHoladay: nice! i was trying to remember that exact thing about enums lol. i don't think i've had the opportunity before to use `as const` but i like it.

BillyFigueroa: > nice! i was trying to remember that exact thing about enums lol. i don't think i've had the opportunity before to use `as const` but i like it.

very neat. I dig it

TristanHoladay (approved):

PR #270, opened 2024-08-30
feat: alert users to cluster disconnection

### LINKED ISSUE #10: Cluster Disconnection Alert
### Description
At the moment, if the engine is connected to a cluster and I kill the cluster the data continues to show in the frontend even upon refreshing the page. I'm guessing this has to do with the resource cache / stores set up. If i create a new cluster and refresh the page, the old data still shows. Killing and restarting the backend with a browser refresh does then show the new cluster data.

### Feature
An alert in the frontend telling users the cluster has been disconnected and a way to connect the backend to the new cluster without killing and restarting.

TristanHoladay:
## Description
Adds a custom watch error handler to the backend to watch for cluster disconnection errors. Upon cluster disconnection the frontend will be notified and pop up an alert in the user's view to let them know. 

### Reconnecting
When a disconnection error is encountered the backend will begin a retry loop to restart the informers with a new k8s client.

## Related Issue

Resolves #10

### DIFF
--- a/pkg/api/resources/list.go
+++ b/pkg/api/resources/list.go
@@ -4,13 +4,19 @@
 package resources
 
 import (
+	"fmt"
+	"io"
 	"strings"
 	"sync"
 
+	"github.com/zarf-dev/zarf/src/pkg/message"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
 )
 
 const (
@@ -53,6 +59,12 @@ func NewResourceList(informer cache.SharedIndexInformer, gvk schema.GroupVersion
 			r.notifyChange(obj, Deleted)
 		},
 	})
+
+	err := informer.SetWatchErrorHandler(r.CustomWatchErrorHandler)
+	if err != nil {
+		// need to talk about bubbling these things up
+		message.WarnErr(err, "failed to set watch error handler")
+	}
 
 	return r
 }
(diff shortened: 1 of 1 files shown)


pkg/api/start.go
```diff
@@ -274,3 +312,58 @@ func serveAuthStatus(w http.ResponseWriter, _ *http.Request) {
...
+type createCache func(ctx context.Context, client *k8s.Clients) (*resources.Cache, error)
+
+func getRetryInterval() time.Duration {
+	if interval, exists := os.LookupEnv("RETRY_INTERVAL_MS"); exists {
+		parsed, err := strconv.Atoi(interval)
+		if err == nil {
+			return time.Duration(parsed) * time.Millisecond
+		}
+	}
+	return 5 * time.Second // Default to 5 seconds if not set
```
UncleGedd, line 333: ❤️

pkg/api/start.go
```diff
@@ -274,3 +312,58 @@ func serveAuthStatus(w http.ResponseWriter, _ *http.Request) {
...
+			k8sResources.Cancel()
+			time.Sleep(getRetryInterval()) // Retry interval
+			k8sClient, err := createClient()
+			if err != nil {
+				fmt.Printf("Retrying to create k8s client: %v\n", err)
+				k8sResources.Client = nil
+				continue
+			}
+
+			// Ccreate a new context and cache
```
decleaver, line 353:
```suggestion
			// Create a new context and cache
```

pkg/api/start.go
```diff
@@ -60,19 +55,40 @@ func Setup(assets *embed.FS) (*chi.Mux, error) {
...
+	}
+
+	cache, err := resources.NewCache(ctx, k8sClient)
+	if err != nil {
+		cancel()
 		return nil, fmt.Errorf("failed to create cache: %w", err)
 	}
 
+	// Create the disconnected channel
+	disconnected := make(chan error)
```
decleaver, line 72:
```suggestion
	disconnected := make(chan error)
	defer close(disconnected)
```

Don't think the channel is being closed

pkg/api/start.go
```diff
@@ -60,19 +55,40 @@ func Setup(assets *embed.FS) (*chi.Mux, error) {
 	r.Use(middleware.Logger)
 	r.Use(middleware.Recoverer)
 
-	ctx := context.Background()
-	cache, err := resources.NewCache(ctx)
+	ctx, cancel := context.WithCancel(context.Background())
```
decleaver, line 58: Think if you do `defer cancel()` here then you can remove it from the two if checks below

pkg/api/start.go
TristanHoladay: looking at this some more and chatting with the machines, i don't think we need to close this channel. the channel needs to stay open for the lifetime of the app, and if the app exits for whatever reason then the channel will be closed anyway. If there's another reason for closing it, then we'll need to close it differently than `defer` so it doesn't close when `Setup()` returns.

decleaver: @TristanHoladay tested locally, closing channel here breaks connection flow

pkg/api/start.go
decleaver: @TristanHoladay tested locally. same here, only want to cancel on error

decleaver (approved):
Comments addressed and issue being created to look into the timing and number of network calls to /health

TristanHoladay:
for multiple health calls https://github.com/defenseunicorns/uds-runtime/issues/299

pkg/api/start_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	k8sResources := &K8sResources{
+		Client:          &k8s.Clients{},
+		Cache:           &resources.Cache{},
+		Cancel:          func() {},
+		OriginalCtx:     "original-context",
+		OriginalCluster: "original-cluster",
+	}
+
+	assert.Nil(t, k8sResources.Client.Clientset)
```
decleaver, line 35: So we've been using `require` instead of `assert` in our tests which stops the tests if the assertion fails.

pkg/api/start_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	disconnected := make(chan error, 1)
+
+	disconnected <- fmt.Errorf("simulated disconnection")
+
+	// Run the handleReconnection function in a goroutine
+	go handleReconnection(disconnected, k8sResources, createClientMock, createCacheMock)
+
+	// Wait for the reconnection logic to complete
+	time.Sleep(200 * time.Millisecond)
```
decleaver, line 139: Did you look into sync.WaitGroup as a way to handle this instead of relying on a sleep?

pkg/api/start.go
```diff
@@ -274,3 +331,69 @@ func serveAuthStatus(w http.ResponseWriter, _ *http.Request) {
...
+	}
+	return 5 * time.Second // Default to 5 seconds if not set
+}
+
+// handleReconnection is a goroutine that handles reconnection to the k8s API
+// passing createClient and createCache instead of callings k8s.NewClient and resources.NewCache for testing purposes
+func handleReconnection(disconnected chan error, k8sResources *K8sResources, createClient createClient,
+	createCache createCache) {
+	for err := range disconnected {
+		fmt.Printf("Disconnected error received: %v\n", err)
```
decleaver, line 360: Since we are already importing `"log"`,  can we use that instead of using `fmt.Print..`?  (Probably other spots we can update too, so i'm fine if this gets updated as part of another issue)

TristanHoladay: that's a good call out. i'll update what i can in this PR.

pkg/api/start_test.go
TristanHoladay: ahh thanks. yeah probably chatGPT copy pasta mistake.

pkg/api/start_test.go
TristanHoladay: hmmm. that's probably a good refactor to make but looking at what the machines say it might take a little more re-working to do. if you're cool with it, i'm going to make an issue to look into it further.

decleaver (dismissed):

decleaver (dismissed):
🚀

decleaver (approved):

PR #276, opened 2024-09-03
fix: api token auth details view

decleaver:
## Description

- Fixes bug accessing details view when API auth is enabled
- Rename http class and file to APIAuth
- Adds a wrapper for api auth checking when creating a new EventSource


UncleGedd (approved):
🚀

PR #278, opened 2024-09-04
chore(ui): redirect to auth page when unauthenticated

decleaver:
## Description
If unauthenticated and you go to a view url directly, you are now redirected to the auth page. Before it would just display an empty table with no indication that you were not authenticated to the api.


BillyFigueroa:
@decleaver the new `unauthenticated.svelte` file should probably be renamed to `component.svelte`. That is the convention set by Jeff

ui/src/lib/features/api-auth/store.ts
```diff
@@ -3,4 +3,4 @@
 import { writable } from 'svelte/store'
 
 export const authenticated = writable(false)
-export const apiAuthEnabled = writable(false)
+export const apiAuthEnabled = writable(<null | boolean>null)
```
TristanHoladay, line 6: i've never seen the typing like this, but the type check isn't failing so.... is there a particular reason for not doing 
```suggestion
export const apiAuthEnabled = writable<null | boolean>(null)
```

BillyFigueroa (approved):

TristanHoladay (approved):

Issue #280, opened 2024-09-04
Handle page refresh and reloads with api token auth


decleaver:
Right now if you reload the app after authenticating, you need to re-authenticate.

This issue is to investigate and implement a better work flow if possible.

PR #282, opened 2024-09-05
refactor(ui): handle page refresh and reloads with api token auth

BillyFigueroa:
## Description
Updating the auth flow to allow for reload of pages and remain authenticated

## Related Issue

- #280


decleaver (dismissed):
tested locally, works like a charm

UncleGedd (approved):
Merging this knowing that the next step is to write a [design doc](https://github.com/defenseunicorns/uds-runtime/issues/294) to discuss and potentially refactor this approach

PR #297, opened 2024-09-09
docs: create a design doc for cluster disconnection handling

TristanHoladay:
## Description
Creates a design document for cluster disconnection handling

## Related Issue

#10 

## Related PR
#270


design-docs/0001-cluster-disconnection.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Changes to Existing Systems:
+
+- Add a new health check route (/health) to monitor the cluster's connection status.
+- Introduce reconnection handling logic in the backend to manage the lifecycle of Kubernetes clients and caches.
+- Wrap route handlers so they dynamically get the latest cache
+
+## Current Problems and Questions
+
+1. By intiating the health check via the frontend call to `/health`, this creates a poll per client. Could this cause strain on the server? Could all these potential error events (for the same cluster disconnection) cause unnecessary reconnection attempts?
```
UncleGedd, line 48: Could this be resolved by polling every 5s instead or constantly?

design-docs/0001-cluster-disconnection.md
```diff
@@ -35,7 +35,7 @@ We will poll the cluster with a server version check. Currently initiating this
 
 ### Frontend Implementation:
 
-When a user lands on the application, triggering the main `src/routes/layout.svelte`, an EventSource is created for `/health` that will now continuously receive updates from the server on cluster connection health. If an error is received, a toast with an extremely long timeout will be displayed to the user. Only a single toast will be added regardless of subsequent error messages. When the connection is restored, the toast is updated to indicate reconnection.
+When a user lands on the application, triggering the main `src/routes/layout.svelte`, an EventSource is created for `/health` that will now continuously receive updates from the server on cluster connection health. If an error is received, a toast with an extremely long timeout will be displayed to the user. Only a single toast will be added regardless of subsequent error messages. When the connection is restored, the toast is updated to indicate reconnection and then removed.
```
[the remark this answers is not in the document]
TristanHoladay: @Madeline-UX requested that the alert stay showing on the page for duration of disconnected status. Should i change this to reflect that idea instead of implementation detail?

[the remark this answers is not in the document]
UncleGedd: I think so, that verbiage makes more sense than "extremely long timeout"

design-docs/0001-cluster-disconnection.md
TristanHoladay: hmm, not sure. i'm thinking that to avoid possible issues here, maybe we should make it such that on startup the poll begins instead of waiting for a client call to start the poll. Then when a client does call to `/health` it gets the latest update (success, error, reconnected) from a channel that the poll has been sending events to. This way even if hundreds of clients (though probably unlikely) are hitting a single instance of the api there will only ever be 1 poll running that can trigger the reconnection handler.

Maybe this is not a right now problem though and should only be considered for future optimizations as needed?

design-docs/0001-cluster-disconnection.md
```diff
@@ -0,0 +1,72 @@
...
+
+**Alternative**
+We will poll the cluster with a server version check. Currently initiating this health check requires the frontend to make a request to `/health` (should this change?). If an error is encountered, the system will emit a disconnection event, triggering the reconnection process in a separate go routine. The reconnection handler will cancel the current cache context (officially stopping the informers), attempt to recreate the Kubernetes client, and reinitialize the cache. This loop will continue until the connection is restored or the application is stopped.
```
decleaver, line 34:
```suggestion
We will poll the cluster with a server health check. Currently initiating this health check requires the frontend to make a request to `/health` (should this change?). If an error is encountered, the system will emit a disconnection event, triggering the reconnection process in a separate go routine. The reconnection handler will cancel the current cache context (officially stopping the informers), attempt to recreate the Kubernetes client, and reinitialize the cache. This loop will continue until the connection is restored or the application is stopped.
```

design-docs/0001-cluster-disconnection.md
UncleGedd: I should be more specific, I was referring to the point about putting strain on the server

TristanHoladay: ahhh, gotcha. currently the poll for cluster health is 30 seconds, so probably not that much strain. maybe that's too long though. I was wondering if that should be configurable. Then it's up to admins to balance speed of check with strain on the server...

design-docs/0001-cluster-disconnection.md
```diff
@@ -0,0 +1,75 @@
...
+- Ensure the system can handle reconnection attempts without causing a complete failure or downtime
+- Keep monitoring the cluster connection state at regular intervals
+
+## Implementation Details
+
+The implementation will consist of the following components:
+
+### Backend Implementation:
+
+**Ideal**:
```
UncleGedd, line 30: For readability, I would refactor `Ideal` and `Alternative` as 2 separate options and then clearly acknowledge which one we are choosing

design-docs/0001-cluster-disconnection.md
```diff
@@ -0,0 +1,75 @@
...
+
+When a user lands on the application, triggering the main `src/routes/layout.svelte`, an EventSource is created for `/health` that will now continuously receive updates from the server on cluster connection health. If an error is received, a toast will be displayed to the user. This error toast should remain on the screen (regardless of user navigation) until a reconnected message is received. Only a single toast will be added regardless of subsequent error messages. When the connection is restored, the toast is updated to indicate reconnection and then removed.
+
+After the reconnection event is received, a custom event will be dispatched to trigger an event listener on the `Datatable` component. This event handler will then restart the table store, which re-fires the eventSource call and gets data from the new cache.
+
+## Changes to Existing Systems:
+
+- Add a new health check route (/health) to monitor the cluster's connection status.
+- Introduce reconnection handling logic in the backend to manage the lifecycle of Kubernetes clients and caches.
+- Wrap route handlers so they dynamically get the latest cache
```
UncleGedd, line 48: > dynamically get the latest cache

Is this because the informers could still have data from the previous connection? If so I'd add that bit of context

UncleGedd (approved):

PR #317, opened 2024-09-11
spike: reload data on cluster reconnection

TristanHoladay:
## Description
This is a throw away draft for exploring reloading data on cluster reconnection.


ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -127,7 +128,14 @@
 
     // Bind the keyboard navigation event
     window.addEventListener('keydown', keyboardNavigate)
-    const stop = rows.start()
+    let stop = rows.start()
+
+    window.addEventListener('clusterReconnected', () => {
```
BillyFigueroa, line 133: Hey Tristan, is there a reason you chose to use the window event listeners instead of the svelte one?

i.e.

```Javascript
import { createEventDispatcher } from 'svelte';

const dispatch = createEventDispatcher();

dispatch('server-reconnect', { text: 'Hello!' });
```

Then in the definition of the table you can do this..

```Javascript
<DataTable {columns} {createStore} {name} {description} on:server-reconnect={handleReconnect} />
```

Finally in the component itself you define a function to handle the event


```Javascript
const handleReconnect = (data) => {
  stop()
  rows = createStore()
  stop = rows.start()
}
```

I will say, your approach is less code, as with this style, we need to add on `on:server-reconnect` to every table we want to reload data

pkg/api/start.go
```diff
@@ -274,3 +331,69 @@ func serveAuthStatus(w http.ResponseWriter, _ *http.Request) {
...
+
+// withLatestCache returns a wrapper lambda function, creating a closure that can dynamically access the latest cache
+func withLatestCache(k8sResources *K8sResources, handler func(cache *resources.Cache) func(w http.ResponseWriter, r *http.Request)) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		handler(k8sResources.Cache)(w, r)
+	}
+}
+
+type createClient func() (*k8s.Clients, error)
+type createCache func(ctx context.Context, client *k8s.Clients) (*resources.Cache, error)
```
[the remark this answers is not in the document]
TristanHoladay: Sure! i should note too that i made this branch off of the `cluster-disconnection` branch. So a lot of these changes enable the reload but really there's only 2 files - `ui/src/lib/utils/cluster-check/cluster-check.ts` and `ui/src/lib/components/k8s/DataTable/component.svelte` - that i changed on this branch for reloading the data on reconnection.

You can think of the cache (`pkg/api/resources/cache.go`) as a big object (a struct in Go) containing resource lists (`pkg/api/resources/list.go`). These lists are populated and updated by informers (a kubernetes client-go feature). This cache is created when the server is started. All resource routes are "bound" to the matching resource list on the cache. So when you hit `/api/v1/resources/workloads/pods` it's accessing the cache.Pods resource list. One benefit of this, but also "gotcha", is that even if the cluster disconnects this cache still exists and the routes are bound to it, meaning users can still see all the data in the cache. They're just not receiving updates anymore. So when we reconnect we have to re-create the cache, which starts new informers to re-create the resource lists. Now users will receive latest data and any new updates.

ui/src/lib/components/k8s/DataTable/component.svelte
TristanHoladay: I see that i can't use `on:server-reconnect` in the Datatable because we can't do custom events on HTML elements. So it has to be on a Svelte component. If i do it when we call `<Datatable ...>` i can't access the handler from within the Datatable component. so then the handler has to be exported say from helpers.ts, but then it doesn't have access to the stop() returned from the initial `rows.start()`. There's probably a way to still get it to all work, but might not be worth it.

I guess we could do this event handler on one of the icon components in the Datatable component, but that's kind of jank lol.

BillyFigueroa: > I see that i can't use `on:server-reconnect` in the Datatable because we can't do custom events on HTML elements. So it has to be on a Svelte component. If i do it when we call `<Datatable ...>` i can't access the handler from within the Datatable component. so then the handler has to be exported say from helpers.ts, but then it doesn't have access to the stop() returned from the initial `rows.start()`. There's probably a way to still get it to all work, but might not be worth it.
> 
> I guess we could do this event handler on one of the icon components in the Datatable component, but that's kind of jank lol.

Ok. Glad we could use the window even handling then!

TristanHoladay:
spike complete

Issue #287, opened 2024-09-06
Confirm status coloring


BillyFigueroa:
### Describe what should be investigated or refactored
Ensure that all columns that are supposed to have a color for status, have the proper status color

PR #318, opened 2024-09-11
feat(ui): confirm status coloring

BillyFigueroa:
## Description
Ensure that all statuses in the various pages have the proper status colors
...

## Related Issue

- #287


ui/src/lib/features/k8s/networks/services/component.test.ts
```diff
@@ -4,6 +4,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import '@testing-library/jest-dom'
 
+import StatusComponent from '$components/k8s/Status/component.svelte'
```
UncleGedd, line 7: IMO having a K8s Service be `Active` doesn't make sense. I don't think they can be in any other status, IMO we can remove this column

BillyFigueroa: > IMO having a K8s Service be `Active` doesn't make sense. I don't think they can be in any other status, IMO we can remove this column

Hey Gedd, are you saying we should remove the status column from the table?

UncleGedd: Yes recommending we remove, noting that neither K9s nor Kubewall show a status field

BillyFigueroa: > Yes recommending we remove, noting that neither K9s nor Kubewall show a status field

Copy. Will push a change asap

UncleGedd (approved):

PR #319, opened 2024-09-11
chore(api): complete api tests for all routes

TristanHoladay:
## Description
This PR finishes out the api testing by ensuring there's a test for each route.

To add resources to the cluster not already created by the deploy-load task, a `load-api.js` script, using KFC, has been added to `hack/load-test/`.


hack/load-test/rc.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: node.k8s.io/v1
```
UncleGedd, line 1: TIL about `RuntimeClass`...

UncleGedd (approved):

Issue #305, opened 2024-09-10
Change initial state of tables to be sorted by 'namespace' (alphabetically)


DannyDTenacious:
### Is your feature request related to a problem? Please describe

Currently, most tables are initially sorted alphabetically by their "name"; however, being sorted by "namespace" (in alphabetical order) would be more helpful as namespace indicates how the individual lines are related to each other (based on recent user feedback interviews). Any tables that are not a 'running log' should follow this pattern for default sorting.

Pages with tables to be sorted by "namespace":
/Applications/Endpoints
/Workloads/Pods
/Workloads/Deployments
/Workloads/DaemonSets
/Workloads/StatefulSets
/Workloads/Jobs
/Workloads/CronJobs
/Configs/UDSPackages
/Configs/UDSExemptions
/Configs/ConfigMaps
/Configs/Secrets
/ClusterOps/HPA
/ClusterOps/PodDisruptionBudges
/ClusterOps/ResourceQuotas
/ClusterOps/LimitRanges
/Networks/Services
/Networks/VirtualServices
/Networks/NetworkPolicies
/Networks/Endpoints
/Storage/PersistentVolumeClaims

PR #326, opened 2024-09-13
refactor(ui): update the default sorting for tables

BillyFigueroa:
## Description
 update the default sorting for tables with namespace columns

## Related Issue

- #305

### DIFF
--- a/ui/src/lib/features/k8s/applications/endpoints/store.ts
+++ b/ui/src/lib/features/k8s/applications/endpoints/store.ts
@@ -57,7 +57,7 @@ export function createStore(): ResourceStoreInterface<Resource, Row> {
       // Remove duplicate URLs
       .filter((e, i, a) => a.findIndex((x) => x.table.url.sort === e.table.url.sort) === i)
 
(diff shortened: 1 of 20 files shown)


UncleGedd (approved):
Thanks for this! Could also make `'namespace'` the default arg for `sortBy` in the constructor, no problem with this approach though

BillyFigueroa:
> Thanks for this! Could also make `'namespace'` the default arg for `sortBy` in the constructor, no problem with this approach though

I would have to then remove the namespace from the places where I set it so that when we don't have any namespace specified it uses the default. If we pass a value in there, it will assign one. If we don't then it uses the default

I can do that if you prefer that. This way is a little more explicit

UncleGedd:
> > Thanks for this! Could also make `'namespace'` the default arg for `sortBy` in the constructor, no problem with this approach though
> 
> I would have to then remove the namespace from the places where I set it so that when we don't have any namespace specified it uses the default. If we pass a value in there, it will assign one. If we don't then it uses the default
> 
> I can do that if you prefer that. This way is a little more explicit

All good, the way you have it is fine

PR #335, opened 2024-09-17
feat(api): implement authsvc in uds package and use jwt in api

### LINKED ISSUE #239: Integrate with Auth Service

- **Given** a state
- **When** an action is taken
- **Then** something happens

(optional) A clear and concise description of any alternative solutions or features you've considered.

TristanHoladay:
## Description
This turns authsvc on via the UDS Package CR and makes use of the JWT for authentication / authorization when running in cluster. For now we will just check that the token is part of the headers. In follow-on PRs, we'll lock routes down based on roles  and groups.

## Related Issue

Resolves #239


UncleGedd (approved):
tested with local dev and local cluster

PR #339, opened 2024-09-17
fix(api): errors from trying to get kubecontext while in cluster

TristanHoladay:
## Description
Fixes getCurrentContext() errors that occur when running the application in the cluster by not running reconnection handling in cluster and serving a different health check response to the UI.

## Testing
Tested manually and confirmed works. Will be submitting a follow up PR with a nightly smoke test to catch this kind of thing. It was a lot of code changes for this PR though.

### DIFF
--- a/pkg/api/start.go
+++ b/pkg/api/start.go
@@ -29,6 +29,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	httpSwagger "github.com/swaggo/http-swagger/v2"
+	"k8s.io/client-go/rest"
 )
 
 type K8sResources struct {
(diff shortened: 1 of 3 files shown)


pkg/api/start.go
```diff
@@ -352,6 +356,12 @@ func getRetryInterval() time.Duration {
 	return 5 * time.Second // Default to 5 seconds if not set
 }
 
+// isRunningInCluster checks if the application is running inside cluster
+func isRunningInCluster() bool {
+	_, err := rest.InClusterConfig()
+	return err == nil
```
UncleGedd, line 362: Docs for `InClusterConfig`:

> // InClusterConfig returns a config object which uses the service account
// kubernetes gives to pods. It's intended for clients that expect to be
// running inside a pod running on kubernetes. It will return ErrNotInCluster
// if called from a process not running in a kubernetes environment.

So instead of checking if any `err == nil`, let's check for `ErrNotInCluster` and handle other errors as well

pkg/api/start.go
```diff
@@ -66,8 +67,19 @@ func Setup(assets *embed.FS) (*chi.Mux, error) {
...
-	go handleReconnection(disconnected, k8sResources, k8s.NewClient, resources.NewCache)
+	// Get current k8s context and start the reconnection goroutine if NOT in cluster
+	if !isRunningInCluster() {
+		currentCtx, currentCluster, err := k8s.GetCurrentContext()
+		if err != nil {
+			k8sResources.Cancel()
+			return nil, fmt.Errorf("failed to get current context: %w", err)
+		}
+
+		k8sResources.OriginalCtx = currentCtx
```
UncleGedd, line 78: What is meant by `OriginalCtx`? Can that be updated to just `ctx` or maybe if you want to be more explicit `currentCxt`?

Also, can we make some of the fields in `K8sResources` private?

TristanHoladay: OriginalCtx and Cluster are for making sure we don't connect to another cluster on reconnection. i imagined this naming would change when we introduce multi cluster switching.

pkg/api/start.go
```diff
@@ -356,10 +362,17 @@ func getRetryInterval() time.Duration {
...
-func isRunningInCluster() bool {
+// isRunningInCluster checks if the application is running in cluster
+func isRunningInCluster() (bool, error) {
 	_, err := rest.InClusterConfig()
-	return err == nil
+
+	if err == rest.ErrNotInCluster {
+		return false, nil
+	} else if err != nil {
+		return true, err
```
UncleGedd, line 372: If there's a an err, recommened we respond with a negative

```suggestion
		return false, err
```

TristanHoladay: i was thinking about that, but i think we still want to affirm if it's in cluster, more particularly for when it's called by the health check handler. if there's an error but it's in cluster then we still want to not send an actual cluster check that requires the kubecontext from local.

PR #340, opened 2024-09-17
fix: historical usage loading

decleaver:
## Description
Fixes bug being seen specifically when looking at larger clusters where the app freezes/crashes if you try to load the cluster overview page  immediately after launching.


UncleGedd (approved):
🚀

PR #331, opened 2024-09-16
chore: api auth state design doc

decleaver:
## Description
Update for considering handling api auth state in backend (using cookies) vs frontend

PR #343, opened 2024-09-18
chore: move api token auth state management to backend using cookies

decleaver:
## Description
This pr updates the way api token auth state management is being handled. Instead of being handles in the frontend using stores and sessionStorage, we are now using session cookies. Design is documented in https://github.com/defenseunicorns/uds-runtime/pull/331


TristanHoladay:
really great job on this refactor. i love how much cleanup happened. UI auth flow is way easier to track to now. The only thing i see that would be nice is some unit tests for session.go.

decleaver:
> really great job on this refactor. i love how much cleanup happened. UI auth flow is way easier to track to now. The only thing i see that would be nice is some unit tests for session.go.

Added ❤️

UncleGedd:
Chatted sync, going to explore refactoring some of the middleware and using the auth middleware for all routes in desktop mode

decleaver:
> Chatted sync, going to explore refactoring some of the middleware and using the auth middleware for all routes in desktop mode

@UncleGedd  refactored where the main middleware code is from auth to middleware. I wasn't able to move the middleware from the `r.With` calls to `r.Use` because  we can't have it apply to every endpoint, because we still need to be able to hit the /api-auth endpoint to be able to do the initial token authentication without being blocked.

I also ended up leaving the auth middleware chain where it was because in order to move it to the middleware package I would end up having to wrap it again in another function since it needs to know the value of apiauth so it ends up just being more code and an additional step when tracing/debugging.

Happy to discuss some more as well

UncleGedd (approved):

Issue #301, opened 2024-09-09
Add sizing to all tables that have cells that need to be truncated


BillyFigueroa:
### Is your feature request related to a problem? Please describe
Add sizing to all tables that have cells that need to be truncated so that the other columns are properly sized

### Describe the solution you'd like

- **Given a table that has table cells where the content is very wide
- **When the table is resized or viewed at a smaller view size
- **Then we should clip the text and add ellipsis `content...` and show the content on hover

#### This is how we add sizes to the tables and truncate them
<img width="379" alt="Screenshot 2024-09-09 at 2 57 52 PM" src="https://github.com/user-attachments/assets/e825fa1d-f3e2-426e-b149-ded7e1799e35">

PR #364, opened 2024-09-20
feat(ui): truncating certain table cells

BillyFigueroa:
## Description
Add sizing to all tables that have cells that need to be truncated so that the other columns are properly sized

## Related Issue

- #301 

## Screenshots
Screenshot 2024-09-20 at 10 28 40 AM
Screenshot 2024-09-20 at 10 28 51 AM
Screenshot 2024-09-20 at 10 29 32 AM

### DIFF
--- a/ui/src/lib/components/k8s/DataTable/component.svelte
+++ b/ui/src/lib/components/k8s/DataTable/component.svelte
@@ -296,7 +296,7 @@
                         </button>
                       {:else if style?.includes('truncate')}
                         <Tooltip title={value}>
-                          <div class={`w-full ${style}`}>
+                          <div class={`w-full ${style}`} style:width="100%">
                             {value}
                           </div>
                         </Tooltip>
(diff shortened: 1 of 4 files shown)


UncleGedd:
Is there a reason why we only truncate certain resources (configmaps and secrets in this case)?

ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -296,7 +296,7 @@
                         </button>
                       {:else if style?.includes('truncate')}
                         <Tooltip title={value}>
-                          <div class={`w-full ${style}`}>
+                          <div class={`w-full ${style}`} style:width="100%">
```
koesbong, line 299: is there a class that sets 100% width so we don't use style?

BillyFigueroa: > is there a class that sets 100% width so we don't use style?

`w-full` right above. I ll remove it. I was switching between dynamic style and the classes. FYI, this PR is on hold. Jeff and Gedd want to figure out if we can do this with no JS and use a CSS only tooltip option. Using a table here complicates layout manipulation with how tables determine widths

Issue #367, opened 2024-09-20
Security Hub integration design doc


UncleGedd:
### Description

UDS Runtime and UDS Marketplace need to data from UDS Security Hub. Today, Security Hub releases a sqlite database in S3 with all of the necessary security data for Runtime and Markplace.

Create a design doc to discuss integration. Key points:
- discuss security hub release cadence and how the sqlite artifact will make it's way into Runtime
- discuss what data is necessary for Runtime and Marketplace views
- discuss how Runtime will query the database and the API layer on top of the database
    - big picture: Runtime hosts the database and has an API for accessing the data, this API will be used by both Runtime and Marketplace to display Security Hub data

decleaver:
https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=2869-4943&node-type=canvas&t=HB6yHbGhSziTyWzd-0

adam-defenseunicorns:
Needs to be revisited after the new Data Schema.

PR #374, opened 2024-09-20
chore(docs): design document for uds security hub integration

TristanHoladay:
## Related Issue

- #367


TristanHoladay:
closing since this will live in internal docs

PR #387, opened 2024-09-24
feat(api): adds jwt group validation when in-cluster

UncleGedd:
## Description

- Performs group validation from the JWT received from Keycloak/Authservice.
- Removes deprecated api auth tests
- In addition to the unit tests, testing was performed by building a local Runtime image + Zarf pkg and deploying into a UDS Core dev cluster, then logging in via Google SSO to verify access

### DIFF
--- a/pkg/api/auth/session.go
+++ b/pkg/api/auth/session.go
@@ -7,7 +7,10 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"net/http"
+	"strings"
 	"sync"
+
+	"github.com/golang-jwt/jwt/v5"
 )
 
 type InMemoryStorage struct {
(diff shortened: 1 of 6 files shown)


pkg/api/auth/session.go
```diff
@@ -101,15 +104,47 @@
...
+			http.Error(w, "Missing Authorization header", http.StatusUnauthorized)
+			return
+		}
 
-		if token == "" {
-			w.WriteHeader(http.StatusUnauthorized)
+		tokenString := strings.TrimPrefix(authHeader, "Bearer ")
+
+		// parse the JWT token without validation (authservice will validate it, we only need the groups here)
+		token, _, err := jwt.NewParser(jwt.WithoutClaimsValidation()).ParseUnverified(tokenString, jwt.Claims(jwt.MapClaims{}))
```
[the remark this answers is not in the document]
UncleGedd: Dismissed because we validate the JWT in Authservice before it hits Runtime

pkg/api/auth/session_test.go
```diff
@@ -40,3 +43,69 @@ func TestRemoveSession(t *testing.T) {
...
+		{
+			name:           "Valid token with allowed group",
+			token:          createToken([]string{"/UDS Core/Admin"}),
+			expectedStatus: http.StatusOK,
+		},
+		{
+			name:           "Valid token without allowed group",
+			token:          createToken([]string{"guest"}),
+			expectedStatus: http.StatusForbidden,
+		},
```
TristanHoladay, line 80: not sure on this, but are we missing a case for a valid token that has no groups? I wasn't sure if the "invalid token" case triggers the err on line 126 of session.go or if it satisfies the a failed check on 132 that falls through to the final error case.

UncleGedd: good call! I'll add a test case

TristanHoladay (approved):

TristanHoladay:
For a follow on, i think we still need to surface the errors to users in the UI. Looks like atm, we're still just console logging when 'eventSource.onerror' occurs for the stores.

PR #389, opened 2024-09-25
chore: update slim core with authsvc deployment and bump k3d version

TristanHoladay:
## Description
This adds authsvc to the slim-core deployment, which allows us to test authsvc protection for nightly and main releases while not bumping our runners to deploy full core. The latest versions of uds-k3d also require an updated version of k3d, so bumped to `5.7.4`.

### DIFF
--- a/tasks/setup.yaml
+++ b/tasks/setup.yaml
@@ -8,11 +8,11 @@ tasks:
       - task: commmon:k3d-full-cluster
 
   - name: slim-cluster
-    description: "Create a k3d cluster and deploy core slim dev with metrics server"
+    description: "Create a k3d cluster and deploy core slim dev with authsvc"
     actions:
       - task: common:k3d-test-cluster
       - task: clone-core
-      - task: metrics-server
+      - task: authsvc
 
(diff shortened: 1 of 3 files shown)


tasks/setup.yaml
```diff
@@ -42,3 +42,9 @@ tasks:
     actions:
       - cmd: uds zarf package create .github/uds-core/src/metrics-server --confirm --flavor upstream -o build/
       - cmd: uds zarf package deploy build/zarf-package-uds-core-metrics-server*.tar.zst --components=metrics-server --confirm
+
+  - name: authsvc
```
UncleGedd, line 46: Can we identify that this task is designed for use only in CI where `.github/uds-core` exists?

Also noting that in another part of these tasks, I'm pretty sure we download uds-core to a `tmp` folder at the root of the repo

UncleGedd: resolved sync

UncleGedd (approved):

UncleGedd (approved):

decleaver (approved):

PR #397, opened 2024-09-27
fix: prevent XSS in AnsiDisplay component

bburky:
## Description

Resolve potential XSS in ansi-to-html. By default, it does not escape `<` or `>` characters, and is unsafe to pass to `insertAdjacentHTML()`.

With `escapeXML: true`, will escape HTML in the input (but will still generate safe unescaped `<span style="...">` as desired)

```
> convert.toHtml('<script>alert(1)</script>')
'&lt;script&gt;alert(1)&lt;/script&gt;'
```

See docs:
https://github.com/rburns/ansi-to-html?tab=readme-ov-file#options

I have not tested this code, I'm just reading through for unsafe code. Please test.


bburky:
Related, if you'd like to catch this issue automatically:

I'd recommend re-enabling `svelte/no-at-html-tags`:
https://github.com/defenseunicorns/uds-runtime/blob/main/ui/eslint.config.js#L38

And maybe add something like the eslint-plugin-no-unsanitized plugin to catch `insertAdjacentHTML()`
https://github.com/mozilla/eslint-plugin-no-unsanitized/

When you're _sure_ you're using these risky raw HTML APIs safely, you can add a comment to document and use an inline `eslint-disable-next-line` to only disable the linter for one line.

bburky:
Related also:
Codeql doesn't support `.svelte` files.

It doesn't seem possible to reconfigure it to support more extensions. Internally it seems that they hardcode file extensions. Their HTML parser does support extracting `<script>` tags, and should work on `.svelte` files.  I temporarily got it working with a hack to rename `*.svelte` files to `*.svelte.html`. You could do that short term if you want to get some partial analysis data, but we should probably file an upstream issue to request support for `.svelte` like they already do for similar formats like `.vue`.

UncleGedd:
Appreciate this @bburky! We'll look into the recommended fixes. Noting that we don't use the `ANSIDisplay` component today, but think it will get used in an upcoming view. Before we actually use it, we will ensure the XSS vuln is patched

UncleGedd (approved):
thank you!

Issue #238, opened 2024-08-21
Improve Pepr implementation efficiency and UX


adam-defenseunicorns:
### Description
Need to investigate and implement optimizations for loading and rendering Pepr data

### Notes
- don't bind every time (use same pattern as observers)
- do Pepr streams on demand, but not forever

![Screenshot 2024-09-26 at 11 12 36 AM](https://github.com/user-attachments/assets/6ee71dad-c584-4fdc-aa8a-2e4b9197a72a)

TristanHoladay:
not sure if this is the right approach, but was wondering about this, and if you can't use the informer factories (regular or dynamic) maybe you can use the underlying `watch` api to create your own: https://stackoverflow.com/questions/40975307/how-to-watch-events-on-a-kubernetes-service-using-its-go-client

UncleGedd:
Closed with a slight improvement (less noticeable on large clusters), but we need to refactor the backend Pepr monitor logic. Tracking in https://github.com/defenseunicorns/uds-runtime/issues/433

PR #402, opened 2024-09-30
feat(api): adds caching layer to pepr endpoints

UncleGedd:
## Description

Adds a simple caching mechanism to the Pepr streams. This makes it so that data is loaded immediately after the user's first visit to the Monitor/Pepr page.

### Details
- Cache is invalidated after 5 minutes or after it exceeds 10MB
- If the cache has data, the data is served followed by the normal Pepr streams

## Related Issue

- https://github.com/defenseunicorns/uds-runtime/issues/238


pkg/api/monitor/cache.go
```diff
@@ -0,0 +1,61 @@
...
+	return &Cache{
+		buffer: nil,
+	}
+}
+
+// Get returns the cached buffer
+func (c *Cache) Get() *bytes.Buffer {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.buffer
```
catsby, line 34: `Get()` here needs to either return a deep copy of `c.buffer`, or the lock needs to be public and the consumer of `Get()` needs to handle locking and unlocking. 

As written, locking during the `Get` here and unlocking on defer will only prevent writes to the buffer during the `Get` call itself. Later in the code (~line 53) there's no lock on the buffer but we read from it (`Bytes()`) and there's nothing to stop something else from getting a lock on the buffer and writing to it at the same time. Because `Set()` completely replaces `c.buffer`, that could be a big problem as we're trying to read from it on line 53

UncleGedd: Good catch, thanks!

pkg/api/monitor/pepr.go
```diff
@@ -72,20 +80,26 @@ func Pepr(w http.ResponseWriter, r *http.Request) {
...
+
 				if err := bufferWriter.Flush(w); err != nil {
 					message.WarnErr(err, "Failed to flush buffer")
 					return
 				}
+
+				// Update the cached buffer if on default stream
+				if streamFilter == "" {
+					tmpCacheBuffer.Write(data)
+					streamCache.Set(tmpCacheBuffer)
```
TristanHoladay, line 101: would it cause issues to move this above the `bufferWriter.Flush(w)...` and pass `bufferWriter.buffer` to `streamCache.Set()`?

decleaver:
❤️ the refactor of code from the page.svelte to helpers.ts

pkg/api/monitor/pepr.go
UncleGedd: My thought there was to ensure we flush the data before writing to the cache. If the cache breaks (for whatever reason) I still wanted core functionality to take precedence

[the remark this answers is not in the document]
UncleGedd: The `tmpCacheBuffer` is what's being pointed to by `streamCache` (the global stream cache). So if we reset the `tmpCacheBuffer` we lose the cache. However, the `tmpCacheBuffer` is cleaned up every time the user leaves the page

decleaver (approved):
:shipit:

Issue #378, opened 2024-09-23
Explore: Localhost TLS like UDS Core for Runtime


adam-defenseunicorns:
### Is your feature request related to a problem? Please describe

> possible considerations around localhost tls things like we do with uds core due to the SSE http limits

The local version of Runtime currently uses `localhost:8080`, which limits us to 6 Server-Sent Events (SSE) connections at a time. This limitation stems from the use of non-TLS connections, which defaults to the HTTP/1 specification. The restricted number of SSE connections constrains our ability to provide real-time updates and limits the scalability of our application, particularly for users who require multiple simultaneous data streams.

### Describe the solution you'd like

- **Given** the current setup using `localhost:8080` for the desktop version of Runtime
- **When** we implement this feature
- **Then** we should consider:
  1. Upgrading `uds ui` to use a domain we own (e.g., `uds.dev` or similar) instead of `localhost`
  2. Implement TLS for all connections to this new domain
  3. Leverage the HTTP/2 specification, which becomes available with TLS
  4. Increase the number of possible SSE connections from 6 to ~ 100
  5. Update all relevant documentation to reflect the new domain and connection method
  6. Implement proper certificate management for the TLS connections

This solution would significantly increase the number of available SSE connections, improving the application's ability to handle real-time updates and enhancing overall scalability.

### Describe alternatives you've considered

1. WebSocket Implementation: Replace SSE with WebSockets, which don't have the same connection limitations. However, this would require more significant changes to both server and client-side code.

2. Polling with Long-Polling Fallback: Implement a polling mechanism with long-polling as a fallback. This could reduce the need for multiple persistent connections but might increase latency and server load.

3. Connection Pooling: Implement a connection pooling mechanism on the server-side to manage and reuse a limited number of connections more efficiently. This could help mitigate the issue without changing the underlying protocol but wouldn't fully solve the connection limit problem.

### Additional context

Implementing this feature would require changes to both the server and client-side code. It would also necessitate the acquisition and management of TLS certificates for the new domain. 

Key benefits include:
- Increased number of simultaneous SSE connections (from 6 to ~100)
- Improved security through TLS encryption
- Better performance and reduced latency due to HTTP/2 features like multiplexing
- Enhanced scalability for future feature additions

PR #405, opened 2024-10-01
feat(api): use tls when running locally

UncleGedd:
## Description

Upgrades our connection to TLS and HTTP/2 when running locally

## Related Issue

- https://github.com/defenseunicorns/uds-runtime/issues/378


hack/certs/README.md
```diff
@@ -0,0 +1,5 @@
+# Certs
+
+The certs in this directory are primarily for used for dev'ing on UDS with HTTPS. They are also being used  to enable TLS when running UDS Runtime locally (such as when doing `uds ui`).
```
TristanHoladay, line 3:
```suggestion
The certs in this directory are primarily used for dev'ing on UDS with HTTPS. They are also being used to enable TLS when running UDS Runtime locally (such as when doing `uds ui`).
```

pkg/api/start.go
```diff
@@ -82,15 +82,15 @@ func Setup(assets *embed.FS) (*chi.Mux, error) {
 	inCluster, err := isRunningInCluster()
```
decleaver, line 82: thoughts on moving this to the top and then just using the inCluster var for all of the returns?

pkg/api/start.go
```diff
@@ -44,10 +44,10 @@ type K8sResources struct {
 // @license.url http://www.apache.org/licenses/LICENSE-2.0.html
 // @BasePath /api/v1
 // @schemes http https
-func Setup(assets *embed.FS) (*chi.Mux, error) {
+func Setup(assets *embed.FS) (*chi.Mux, bool, error) {
```
decleaver, line 47: thoughts on adding comment to this function explaining what the boolean value returned represents (incluster or not)

main.go
```diff
@@ -13,18 +14,51 @@ import (
...
+var localCert []byte
+
+//go:embed hack/certs/key.pem
+var localKey []byte
+
 func main() {
 	message.SetLogLevel(message.DebugLevel)
-	r, err := api.Setup(&assets)
+
+	r, inCluster, err := api.Setup(&assets)
```
TristanHoladay, line 26: what do you think about not returning inCluster but instead make isRunningInCluster() public and call directly? (I'm moving that method into a different pkg fysa in my refactor PR). I think it would be nice to not return more things from Setup() and would make more sense at the top level as to what is meant by "inCluster".

ui/playwright.config.ts
```diff
@@ -2,12 +2,12 @@ import { defineConfig } from '@playwright/test'
 import { loadEnv } from 'vite'
 
 const { VITE_PORT_ENV } = loadEnv('dev', process.cwd())
-const port = VITE_PORT_ENV ?? '8080'
+const port = VITE_PORT_ENV ?? '8443'
```
TristanHoladay, line 5: would you mind leaving a small comment here so it's easy for people, not familiar with the server code, to understand why `8443` instead of the typical `8080`.

pkg/api/start.go
UncleGedd: Done!

pkg/api/start.go
UncleGedd: done!

main.go
UncleGedd: discussed sync; going to move `isRunningInCluster` up for now and plan on refactoring in the next tickets

ui/playwright.config.ts
UncleGedd: added!

decleaver (approved):
multiple tabs!!!

decleaver (approved):
dismissed

PR #406, opened 2024-10-01
chore(api): refactor organization of reconnection logic

TristanHoladay:
## Description
Move the reconnection code out of the api package and into its own package for maintainability.


pkg/api/start.go
```diff
@@ -50,16 +37,24 @@ func Setup(assets *embed.FS) (*chi.Mux, bool, error) {
 	var apiAuth bool
 	var token string
 
-	inCluster, err := isRunningInCluster()
+	// Create a k8s session {Clients, Cache, CurrentCtx, CurrentCluster, Cancel, InCluster}
+	k8sSession, err := session.CreateK8sSession()
```
UncleGedd, line 41: Love the naming! Let's remove `{ ... }` from the comment and do an err check before `inCluster := k8sSession.InCluster`

pkg/k8s/session/session.go
```diff
@@ -0,0 +1,124 @@
...
+		CurrentCtx:     currentCtx,
+		CurrentCluster: currentCluster,
+		Cancel:         cancel,
+		InCluster:      inCluster,
+	}
+
+	return session, nil
+}
+
+// handleReconnection is a goroutine that handles reconnection to the k8s API
```
UncleGedd, line 69:
```suggestion
// HandleReconnection is a goroutine that handles reconnection to the k8s API
```

pkg/api/start.go
```diff
@@ -358,80 +307,8 @@ func checkForClusterAuth() bool {
...
-				continue
-			}
-
-			k8sResources.client = k8sClient
-			k8sResources.cache = cache
-			k8sResources.cancel = cancel
-			log.Println("Successfully reconnected to k8s and recreated cache")
-			break
-		}
+		handler(k8sSessionCTX.Cache)(w, r)
```
UncleGedd, line 312: Let's use `k8sSessionCtx` instead of `k8sSessionCTX`

UncleGedd: nah `k8sSession` is better, I thought that was literally a `context` by the naming

BillyFigueroa:
Nice refactor @TristanHoladay. I m still too new to go to feel comfortable approving this but looks good to me

pkg/k8s/session/session.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		}
+
+		// If running in cluster don't check for version and send error or reconnected events
+		if k8sSession.InCluster {
+			checkCluster = func() {
+				response := map[string]string{
+					"success": "in-cluster",
+				}
+				data, _ := json.Marshal(response)
```
decleaver, line 181: Can you add an error check here, think we check in most places we do a Marshal

decleaver (approved):
:shipit:

Issue #376, opened 2024-09-23
Graceful handling of unavailable K8s CRDs


adam-defenseunicorns:
### Is your feature request related to a problem? Please describe

When deploying applications that depend on Custom Resource Definitions (CRDs), the absence of these CRDs can lead to deployment failures and errors. Currently, the system does not gracefully handle situations where required CRDs are missing.

### Describe the solution you'd like

- **Given** a cluster where certain CRDs are not available
- **When** an application that depends on these CRDs is deployed
- **Then** Runtime should...
  - Detect the missing CRDs
  - Log warnings about the unavailable CRDs
  - Offer clear error messages and documentation on how to resolve CRD-related issues
  - **_Other/More??_**

### Describe alternatives you've considered

1. Pre-deployment CRD check: Implement a pre-deployment step that verifies all required CRDs are present before initiating the deployment process. Do we already do any of this in CLI and if not would this be worth looking into?
2. Automatic CRD installation: Automatically install missing CRDs as part of the deployment process. This could potentially solve the issue but might not be desirable in all environments due to security or governance concerns.

### Additional context

This feature would be particularly useful for organizations managing multiple Kubernetes clusters with varying CRD availability, such as development, staging, and production environments. It would also benefit teams working on applications that use optional features dependent on specific CRDs, allowing for more flexible deployments based on the target environment's capabilities.

decleaver:
Currently if there is a missing CRD you get logs similar to the following:
`E0930 08:40:17.371313   40214 reflector.go:158] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watch uds.dev/v1alpha1, Resource=exemptions: failed to list uds.dev/v1alpha1, Resource=exemptions: the server could not find the requested resource" logger="UnhandledError"`
and  the table displays "No resources found". Not sure what is desired.
![image](https://github.com/user-attachments/assets/beaa4ce0-72d0-403c-b1d9-cfd7188d6d8c)

PR #408, opened 2024-10-01
chore: implementation for handling crds that dont exist

decleaver:
## Description
Adds additional error logging specifying that a CRD doesn't exist in the cluster and also displays the same information in the table.

Example Logs:
image

Example of table view:
image

vs CRD exists and no resources
image

## Related Issue

- #376


UncleGedd:
Good start! I left a couple of comments on the design doc regarding expected UX and recovery. TLDR: if something is wrong we should tell the user how to fix it, and enable seamless recovery once they install the CRD 

Is it possible to make this solution a bit more generic? For example, the code as of today works for detecting missing UDS package resources but UDS Exemptions doesn't show the error message.

Screenshot 2024-10-02 at 9 06 45 AM

decleaver:
> Good start! I left a couple of comments on the design doc regarding expected UX and recovery. TLDR: if something is wrong we should tell the user how to fix it, and enable seamless recovery once they install the CRD
> 
> Is it possible to make this solution a bit more generic? For example, the code as of today works for detecting missing UDS package resources but UDS Exemptions doesn't show the error message.
> 
> Screenshot 2024-10-02 at 9 06 45 AM
> 
> By "more generic" I mean instead of applying it to specific resources, can we apply to it all CRDs?

Yeah, thats a bug... Should be currently applying to the 3 dynamic informer resources (virtualservices, uds packages and uds exemptions). But yeah, can update it so it applies to all CRDs. Will actually probably make the code a little cleaner.

ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -257,6 +257,15 @@
               <tr class="!pointer-events-none !border-b-0">
                 <td class="text-center" colspan="9">No resources found</td>
               </tr>
+              <!--This case only happens when returning an error when a CRD does not exist in the cluster-->
+            {:else if $rows.length > 0 && typeof $rows[0].resource === 'string'}
```
TristanHoladay, line 261:
```suggestion
            {:else if typeof $rows[0].resource === 'string'}
```

decleaver: think this will throw an error if rows is empty

pkg/api/rest/bind.go
```diff
@@ -11,73 +11,85 @@ import (
...
-		if fields != "" {
-			fieldsList = strings.Split(fields, ",")
-			// If fields are specified, dense data retrieval is required
-			dense = true
-		}
+func handleRequest(w http.ResponseWriter, r *http.Request, resource *resources.ResourceList) {
+	// If true, send full resource data
+	// By default, send the data as a sparse stream
+	dense := r.URL.Query().Get("dense") == "true"
+	// Get the namespace from the URL query
```
UncleGedd, line 18: These comments don't seem to fit with the comment style in the rest of the codebase, for example:
```
	// Get the UID
	uid := chi.URLParam(r, "uid")
```
We don't typically comment every line. Generally speaking, comment should be "higher level" unless there's a weird implementation detail or there is nuance involved. In the example above, it's clear from the code that we're getting a param called from the URL

pkg/api/resources/cache.go
```diff
@@ -78,6 +79,9 @@ type Cache struct {
 	// Metrics
 	PodMetrics     *PodMetrics
 	MetricsChanges chan struct{}
+
+	// CustomResrouceDefinitions
```
UncleGedd, line 83:
```suggestion
	// CustomResourceDefinitions
```

pkg/api/resources/crds.go
```diff
@@ -0,0 +1,116 @@
...
+	definitions map[string]struct{}
+}
+
+func NewCRDs() *CRDs {
+	return &CRDs{
+		definitions: make(map[string]struct{}),
+	}
+}
+
+func (c *Cache) setupCRDInformer() {
```
UncleGedd, line 26: Since this is a `Cache` receiver method, does it make more sense to put this in `cache.go`?

pkg/api/rest/bind.go
decleaver: cleaned up comments

pkg/api/resources/crds.go
decleaver: I dont have strong feelings on it.  I tried putting all the CRD specific functions in this file. Currently we have metrics specific Cache receiver functions in metrics.go as well. I can move these all to Cache.go if you like.

pkg/api/resources/cache.go
decleaver: had to read this more times than i would like to admit to figure out what the difference was lol

pkg/api/resources/crds.go
decleaver: saving for refactor friday :)

TristanHoladay:
tested:
- [x] No CRDs
- [x] Add CRDs
- [x] Add CRs
- [x] Delete CRs
- [x] Delete CRDs

LGTM!

ui/src/lib/components/k8s/DataTable/component.svelte
decleaver: ah yeah!

TristanHoladay: removed the length check and re-compiled:

Screenshot from 2024-10-07 13-58-38

TristanHoladay (approved):

UncleGedd (approved):
great work!

PR #412, opened 2024-10-02
fix: reconnection handling after introducing TLS

TristanHoladay:
## Description
This PR refactors the cluster check and reconnection handling logic that has been affected (on certain systems) by the upgrade to TLS when running locally. See the [Design Doc](https://docs.google.com/document/d/1HBViN2FydUKNRhHkJeBK3utqUft93bJxkPBm3cKnxZA) for more details on the actual errors and why these refactors were needed. 

### Backend
1. Cluster checking has been moved out of the `/health` SSE handlers
This means a single go routine for running a cluster check on a 10 second interval
1. `HandleReconnection()` no longer runs in a separate go routine, listening for errors to the `disconnected` channel, but is called directly by `StartClusterMonitoring()` in the event of an error.
1. `/health` SSE handlers listen for cluster check statuses on the `Status` channel

### Frontend
1. Removed `clusterCheck.onerror`
This prevents 2 things: 1) showing users an error notification when the cluster hasn't actually disconnected, the SSE stream has just been canceled for some network reason; 2) the browser not retrying the `/health` connection when the SSE stream has been canceled.
1. Now determines reconnected status if there was an error msg event followed by a success msg event

## Testing
Manually tested on:
1. Linux host with Brave
a. dev mode (no auth and connections are HTTP/1.1)
b. as compiled binary with TLS and auth enabled
c. as package in uds-core-slim-dev cluster with authservice running
1. Mac host with Brave
a. as compiled binary with TLS and auth enabled

Screenshot from 2024-10-07 13-12-38

Screenshot from 2024-10-07 13-12-58

Screenshot from 2024-10-07 13-13-03


decleaver:
is it possible to move the "no connection" toast? Right now it blocks the namespace selection keeping you from being able to filter by namespace. (im ok with this being a separate issue as well)

TristanHoladay:
yeah i think we can move it easily enough. maybe just up a bit so that you can select namespace but it's still in the general spot Design previously approved.

pkg/k8s/session/session.go
```diff
@@ -70,68 +69,84 @@ func CreateK8sSession() (*K8sSession, error) {
...
+			ks.HandleReconnection()
+		} else {
+			log.Println("Cluster check successful")
+			ks.Status <- "success"
+		}
+	}
+}
+
 // HandleReconnection is a goroutine that handles reconnection to the k8s API
 // passing createClient and createCache instead of calling clients.NewClient and resources.NewCache for testing purposes
```
UncleGedd, line 104: quick fix: need to update this comment

tasks/test.yaml
```diff
@@ -85,12 +85,12 @@ tasks:
         description: deploy minio to test PVCs
 
   - name: smoke
-    description: "runtime nightly smoke tests"
+    description: "runtime nightly smoke tests -- (requires DIR to be set)"
```
UncleGedd, line 88: > -- (requires DIR to be set)

Why does this require DIR to be set?

ui/src/lib/utils/cluster-check/cluster-check.ts
```diff
@@ -1,22 +1,42 @@
-import { get } from 'svelte/store'
-
 import { addToast } from '$features/toast'
 import { toast } from '$features/toast/store'
 
 // checkClusterConnection checks the connection to the cluster and
```
UncleGedd, line 4: old comments?

ui/src/lib/utils/cluster-check/cluster-check.ts
```diff
@@ -1,22 +1,42 @@
...
+    this.#clusterCheck = new EventSource('/health')
+    this.#clusterCheck.onmessage = (msg) => {
+      this.#handleCloseEvt(msg.data)
+      this.#handleDisconnectedEvt(msg.data)
+      this.#handleReconnectionEvt(msg.data)
+    }
+  }
+
+  #handleDisconnectedEvt(data: string) {
+    console.log(data, 'error')
```
UncleGedd, line 22: looks like we are console logging before we know if it's an error

ui/src/lib/utils/cluster-check/cluster-check.ts
```diff
@@ -1,22 +1,42 @@
...
-    const errToast = get(toast).find((t) => t.message === disconnectedMsg)
+export class ClusterCheck {
+  #clusterCheck: EventSource
+  #disconnectedMsg = 'Cluster health check failed: no connection'
+  #disconnected = false
+
+  constructor() {
+    this.#clusterCheck = new EventSource('/health')
+    this.#clusterCheck.onmessage = (msg) => {
+      this.#handleCloseEvt(msg.data)
```
UncleGedd, line 15: Personally I think it's cleaner to lift the message checks (ie. `data === 'error'`, etc) outside of the handlers and into the caller

pkg/k8s/session/session.go
```diff
@@ -175,17 +171,12 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
 
 		// If running in cluster don't check connection
 		if ks.InCluster {
-			sendInClusterStatus(w, flusher)
-			return
+			fmt.Fprintf(w, "data: %s-close\n\n", "success")
```
UncleGedd, line 174: Why are we sending "success-close"? Even so I don't think we need the `%s`

tasks/test.yaml
TristanHoladay: oh shoot i added that before refactoring the DIR to be hardcoded.

ui/src/lib/utils/cluster-check/cluster-check.ts
TristanHoladay: ahh my bad debugging log. i'll remove

pkg/k8s/session/session.go
TristanHoladay: yeah good call in the `%s`. i'm sending close because i'm trying to close the stream since we're in cluster. Ideally we could do it more like `fmt.Fprintf(w, "event: close\ndata: %s-close\n\n", "success")` as the machines indicate, but that does not trigger the onmessage event and I'm not sure if it's actually closing the stream. Closing the request from the backend `r.Context().Done()` was causing runtime errors, so that was a no go.

UncleGedd: where on the frontend are we processing "success-close"?

BillyFigueroa:
Looks good to me, pulled locally and I see the messages appear when expected

pkg/k8s/session/session.go
TristanHoladay: so I was just looking for if close is in the returned message string in cluster-check.ts, but i just pushed a change that should look for a custom close event instead of looking for that string.

pkg/k8s/session/session.go
```diff
@@ -171,7 +171,7 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
 
 		// If running in cluster don't check connection
 		if ks.InCluster {
-			fmt.Fprintf(w, "data: %s-close\n\n", "success")
+			fmt.Fprint(w, "event: close\ndata: in-cluster\n\n")
```
UncleGedd, line 174: Trying to decipher what's happening here. Are we pushing a DOM event from the backend? Also SSE events must start with `data: `

TristanHoladay: yep, that's my understanding that we're changing the event type. When you run it in cluster, if you look at the network calls you'll see that the type under "EventStream" is no longer message, but "close". Since we're closing the stream in this case i think it's ok that data is not first.

Screenshot from 2024-10-08 10-26-02

It looks like in cluster is effected by the network change. I was pointed at k3d-uds and then deleted and recreated k3d-runtime on the side. It triggered stream cancels and retries, but /health didn't retry cause the stream had been closed by that point, which i think is good. Also, when the streams reconnect everything is good in the UI.

Screenshot from 2024-10-08 10-28-19

UncleGedd: gotcha, okay TIL about [named events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#closing_event_streams)

UncleGedd:
Overall great refactor and this seems to have solved the issue (was this actually a TLS issue though?). There's def a lot of complexity here but I think we've spend enough time on it for now

UncleGedd (approved):

BillyFigueroa (approved):

PR #418, opened 2024-10-08
feat(ci): publish the runtime ui build as a release artifact

### LINKED ISSUE #417: Refactor `uds ui`

Don't embed `uds-runtime` into UDS CLI as this will be an issue for certain security scanners

Something like:
- Publish the Runtime frontend as a build / release artifact
- Vendor Runtime in CLI like we do with Zarf + Maru
- Embed the published frontend artifact(s) in CLI (don't forget to ensure the version matches the vendored Runtime version!)
- ???
- Profit

TristanHoladay:
## Description
Publish the ui build as a release artifact so that `uds-cli` can consume it as part of the refactor to vendoring runtime instead of using the runtime binaries.

Also, extracts the listenAndServer logic into a separate function that is called in `main()`. This will allow us in `uds-cli` to call Setup() and then Serve().

## Related Issue
 #417


UncleGedd (approved):
good refactor too!

PR #423, opened 2024-10-09
feat(ui): view all cluster CRDs

### LINKED ISSUE #404: View of CRDs in cluster (with versions)

In my cluster I have a lot of CRDs from different tools (Istio, Prometheus, ...). It would be nice if I had a way to view all CRDs installed as well as the API Versions.

Provide a new page/view on runtime that has all CRDs installed in the cluster, with clear information about the APIVersions. 

It would also be nice if there were an easy way to click a CRD and see all the custom resources in cluster based on that CRD (i.e. if I click `ServiceMonitor` CRD it shows all `ServiceMonitor` resources in my cluster, but this is likely a larger effort to build dynamic views based on CRDs - I would be more interested in the initial addition of a CRD list.

Example of something similar from K9s:
Screenshot 2024-09-30 at 1 54 12 PM

Relevant fields:
- Name
- Group
- Kind (often similar/same as name, but with casing and singular)
- Versions
- Scope (cluster vs namespace)

This may not be a common SRE task, in which case maybe I should just switch to kubectl via CLI or another tool to view this info?

TristanHoladay:
## Description
Adds a view for seeing all CRDs installed in the cluster.

## Related Issue

Resolves #404 

## Comparisons

OpenLens
Screenshot from 2024-10-10 09-55-14

K9s
Screenshot from 2024-10-10 09-56-36

## Runtime
Screenshot from 2024-10-10 14-32-27

Screenshot from 2024-10-10 14-36-44

Screenshot from 2024-10-10 14-36-50


TristanHoladay:
found that it's possible to add multiple event handlers to informers for add, update, delete. Going to try that route for using 1 informer for CRD view and determining if certain CRDs are not in cluster (UDS Package, Exemptions, Virtual Services).

ui/src/lib/features/k8s/crds/component.svelte
```diff
@@ -0,0 +1,16 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
```
UncleGedd, line 1: update license

ui/src/lib/features/k8s/crds/component.test.ts
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
```
UncleGedd, line 1: update license

ui/src/lib/features/k8s/crds/store.ts
```diff
@@ -0,0 +1,33 @@
+import type { V1CustomResourceDefinition as Resource } from '@kubernetes/client-node'
```
UncleGedd, line 1: add license

ui/src/routes/(resources)/custom-resources/[[uid]]/+page.svelte
```diff
@@ -0,0 +1,5 @@
+<script>
```
Comment (UncleGedd) [inline ui/src/routes/(resources)/custom-resources/[[uid]]/+page.svelte:1]: add license

UncleGedd:
Rename and/or add a comment to `addExistsListeners` in `crds.go` to better indicate what that function is doing (and what feature it's supporting)

TristanHoladay:
after refactoring out `UDSCRDs` and a lot of `crd.go`, i manually tested:

- [x] no CRD displays CRD not found error text
- [x] add the CRD, refresh the event stream, CRD is found but no resources
- [x] add a resource and it automatically shows up
- [x] remove the resource and it automatically goes back to "no resources found"
- [x] remove the CRD and it automatically goes back  no CRD error text

pkg/api/resources/cache_test.go
```diff
@@ -53,15 +77,19 @@ func TestBindCoreResources(t *testing.T) {
...
 	}(ctx)
 
 	c.factory.WaitForCacheSync(c.stopper)
+	c.dynamicFactory.WaitForCacheSync(c.stopper)
 
 	// wait for the context to be done
 	<-ctx.Done()
 	defer close(c.stopper)
 
+	// log.Println("got here")
```
UncleGedd, line 90:
```suggestion
```

UncleGedd:
Looks good! One question: we we add new endpoints shouldn't this trigger a update to our Swagger docs? If so, we need to fix our CI to detect the change

TristanHoladay:
good call out! looks like the check was looking for the old docs location in the `git status -s` output. updating that to look for the actual file `pkg/api/docs/docs.go`.

UncleGedd (approved):
Looks great! Noting that we'll follow up in another PR to remove "Events" from the details tab

PR #425, opened 2024-10-10
fix: storageclass view data

decleaver:
## Description
Fixes missing data fields in storageclass view


UncleGedd (changes requested):
instead of `dense=true` can we use field selectors?

decleaver:
> instead of `dense=true` can we use field selectors?

So switching to dense gives us 3 more fields, including the 2 we were missing.

UncleGedd (approved):

PR #426, opened 2024-10-10
chore: refactor auth logic

UncleGedd:
## Description

Refactors auth logic
- consolidates logic for auth outside of `Setup` function
- renames vars/structs related to browser session
- refactors and conslidates in-cluster vs local auth middleware
- migrates design docs to Google Drive
- adds refresh + new tab tests


pkg/api/auth/configure.go
```diff
@@ -0,0 +1,67 @@
...
+
+	config.LocalAuthEnabled = localAuthEnabled
+	if localAuthEnabled {
+		slog.Info("Local auth enabled")
+		token, err := randomString(96)
+		if err != nil {
+			slog.Error("Failed to generate local auth token")
+			os.Exit(1)
+		}
+		config.LocalAuthToken = token
```
TristanHoladay, line 33: putting the token in the config feels a little odd. It might also be odd to put it on the `BrowserSession`, but I'm wondering if that would make more sense.

UncleGedd: good point, it's not really a config, I'll think of a better place to put it

decleaver:
Thanks for copying the design docs over to gdrive, I like the code refactor as well.

Seeing issues when doing a browser refresh or trying to access runtime in a new tab. Getting redirected to "Could not authenticate..." page

PR #447, opened 2024-10-11
chore(ci): e2e test for cluster disconnection and reconnection

### LINKED ISSUE #430: Expand E2E testing for cluster reconnect feature

Ensure we are testing the cluster reconnect feature with live clusters

TristanHoladay:
## Description
E2E tests, plus some minor refactors, for cluster disconnection handling.

## Related Issue

Resolves #430


UncleGedd (dismissed):
Looks great! Noting that we'll follow up in another PR to remove "Events" from the details tab

ui/tests/reconnect.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { expect, test } from '@playwright/test'
+import { K8s, kind } from 'kubernetes-fluent-client'
+
+// Utility function to run shell commands
+function execCommand(command: string) {
+  try {
+    execSync(command, { stdio: 'inherit' })
+  } catch (error) {
+    console.error(`Error running command: ${command}`)
```
koesbong, line 11: do we want the `console.error` here?

pkg/k8s/session/session.go
```diff
@@ -141,8 +151,11 @@ func (ks *K8sSession) HandleReconnection() {
 		ks.Clients = k8sClient
 		ks.Cache = cache
 		ks.Cancel = cancel
-
 		ks.ready = true
+
+		// immediately send success status to client now that cache is recreated
+		ks.Status <- "success"
+		lastStatus = "success"
 		log.Println("Successfully reconnected to cluster and recreated cache")
```
koesbong, line 159: do we need this `log` here too?

ui/tests/reconnect.spec.ts
TristanHoladay: ahh yeah, we probably don't even need to `console.log` there. Mainly just want to catch command errors.

pkg/k8s/session/session.go
TristanHoladay: that's a good question. maybe not, though i have used it for debugging in the last couple weeks so it might be worth keeping. but i think it brings up a bigger team question of where we do and don't want to log.

TristanHoladay:
didn't want to break the reconnection test into a separate test task but stopping and starting the cluster with pepr dev deployed throws errors. I tried putting all configs into a single `defineConfig()` in the `playwright.config.ts` but they get merged, the last overwriting those before it.

pkg/k8s/session/session.go
```diff
@@ -79,24 +80,33 @@ func CreateK8sSession() (*K8sSession, error) {
...
+	_, err := ks.Clients.Clientset.ServerVersion()
+	if err != nil {
+		ks.Status <- "error"
+		lastStatus = "error"
+		ks.HandleReconnection()
+	} else {
+		ks.Status <- "success"
+		lastStatus = "success"
+	}
+}
```
catsby, line 94: For what it's worth, this function seems overloaded or misnamed. A "ping" checks if something is available and reports back; success or failure, with failure probably determined by a timeout or lack of acknowledgement. Our comment calls it a health check though.

If there's any kind of error, we then go to `ks.HandleReconnection()` which is an infinite loop that wont quit until the reconnection is established, or entire process is shut down. Should we bail out after a certain timeout?

catsby (commented):
minor comments

pkg/k8s/session/session.go
```diff
@@ -33,6 +33,8 @@ type K8sSession struct {
 type createClient func() (*client.Clients, error)
 type createCache func(ctx context.Context, client *client.Clients) (*resources.Cache, error)
 
+var lastStatus string
```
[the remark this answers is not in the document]
TristanHoladay: yeah, i added it back in (used to be part of the `K8sSession` struct, which felt odd) because i didn't like that I was doing the server check directly in `ServeConnStatus()` on initial request. This way the client can get an immediate response of the last known status and then wait for any further updates. That can be helpful primarily for cutting down the time it take for the client to receive an error when the SSE streams are canceled and retried by the browser now that we're using HTTP2.

pkg/k8s/session/session.go
TristanHoladay: that's good feedback. i'll work on the naming since i want to reuse this logic for the immediate check and then when the ticker triggers the next one.

I've wondered too about the infinite loop and if it should ever be stopped. this will never matter for in-cluster, so just local deployment, in which case i've been thinking it's more likely the app will not be long running (maybe a bad assumption though). If we do kill the loop after a certain period of time, and then the cluster does become available again, as-is the app would have to be restarted to create the client and cache again. Not saying that means we shouldn't, but just a consideration. Perhaps it would make sense to do some kind of check in the SSE Handler(), so if the user clicks around ,after the reconnection handler has timedout, and it sees there's no cache it tries to kick off the reconection handler again ...

ui/package.json
```diff
@@ -13,6 +13,7 @@
     "lint": "prettier --check . && eslint .",
     "format": "prettier --write .",
     "test:local-auth": "playwright test tests/local-auth.spec.ts --config=playwright.config.local-auth.ts",
+    "test:reconnect": "playwright test tests/reconnect.spec.ts --config=playwright.reconnect.config.ts",
```
UncleGedd, line 16: Do you need to exclude this test in the `playwright.config.ts`? I would think it'd fail when running under `npm run test:integration` 🤔

UncleGedd (dismissed):
test looks great!

TristanHoladay:
Noting that @catsby brought up good points for us to look into, which we'll roll into our planned refactor of the disconnection / reconnection feature.

UncleGedd (approved):

PR #451, opened 2024-10-15
chore: add e2e in cluster testing

decleaver:
## Description

- add e2e testing running runtime in cluster


tasks/build.yaml
```diff
@@ -87,11 +87,38 @@ tasks:
       - cmd: ./uds zarf p publish build/zarf-package-uds-runtime-arm64-${REF}.tar.zst oci://ghcr.io/defenseunicorns/packages/uds
         dir: ${DIR}
 
-  - name: smoke-img-pkg
+  - name: smoke-img-pkg-amd
```
UncleGedd, line 90: casually looking around and understood this PR is still in draft. Is splitting out `arm` and `amd` tasks necessary here? Does a [multi-arch build](https://github.com/defenseunicorns/uds-runtime/blob/010a32024d252fdf419654823fe5961e624948ef/tasks/build.yaml#L72) work here?

ui/playwright.config.incluster.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from '@playwright/test'
+import { loadEnv } from 'vite'
+
+const { VITE_PORT_ENV } = loadEnv('dev', process.cwd())
+
+// use port 8443 because by default we use TLS when running locally
+const port = VITE_PORT_ENV ?? '8080'
+const protocol = 'http'
+const host = 'localhost'
```
UncleGedd, line 9: Again casually looking around, I figured we'd do something like:
- Deploy uds-runtime in cluster (typical slim-dev setup), probably on the tenant gateway
- connect to `runtime.uds.dev` with playwright
- $$$

tasks/build.yaml
decleaver: Dang, this won't work.

_The error message ERROR: docker exporter does not currently support exporting manifest lists in your GitHub workflow typically occurs when using BuildKit to build multi-platform Docker images but trying to export the image to the Docker daemon, which doesn’t support manifest lists (multi-architecture images) by default._

Looks like the options are to either push the multi-arch to an external docker registry or just use single platform build

UncleGedd: ah dang. Ok single platform build it is, appreciate you giving it a shot!

ui/playwright.config.incluster.ts
decleaver: lets chat on this one. Right now I have an ingress resource set up using trafik and a networkpolicy to allow the playwright tests to hit the app.

decleaver: updating to use slimdev & layers

.github/workflows/nightly-release.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: ./.github/actions/setup
 
       - name: smoke-test
-        run: uds run test:smoke --set DIR=hack/smoke-test
+        run: uds run test:smoke --set DIR=hack/test
```
TristanHoladay, line 41: i think we can get rid of these --sets now. I probably forgot to remove them when we hard-coded the DIR in the actual task.

tasks/test.yaml
```diff
@@ -19,10 +19,68 @@ tasks:
...
+          helm repo add istio https://istio-release.storage.googleapis.com/charts
+          helm upgrade -i istio-base istio/base -n istio-system --set defaultRevision=default --create-namespace
+        description: install Istio CRDs
+      - cmd: |
+          helm repo add minio https://charts.min.io/
+          helm upgrade -i -n uds-dev-stack minio minio/minio --create-namespace --set replicas=1 --set mode=standalone --set persistence.size=1Gi \
+            --set resources.requests.memory=256Mi --set resources.requests.cpu=100m
+        description: deploy minio to test PVCs
+
+  - name: deploy-podinfo-amd
```
TristanHoladay, line 52: having the tasks broken out for amd and arm seems totally reasonable and good. if you have time, though, it might be worth looking at turning these into tasks with `inputs`.

ui/tests/table.spec.ts
```diff
@@ -11,33 +13,57 @@ test.describe('DataTable', async () => {
   test('filters rows when we click the namespace link in a row', async ({ page }) => {
     await page.getByRole('button', { name: 'podinfo' }).last().click()
 
-    expect(await page.getByTestId('table-header-results').textContent()).toBe('(showing 1 of 8)')
+    if (isInCluster) {
```
UncleGedd, line 16: Instead of conflating in-cluster and not-in-cluster scenarios, can we just grab the latter number and ensure it's greater than or equal to a number? Even better if we can refactor the existing tests to use the functional layers, I think that'll make the pod numbers line up better

decleaver: yeah, I can update to do greater than or equal. Even with functional layers, the numbers still won't add up because in this case uds-runtime is in the cluster which will throw off the pod numbers

chart/values.yaml
```diff
@@ -1,6 +1,10 @@
 # Copyright 2024 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
+# Test environment flag
+test:
```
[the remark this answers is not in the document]
decleaver: good catch, forgot to clean up

ui/playwright.config.incluster.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from '@playwright/test'
```
TristanHoladay, line 1: Just found that you can make multiple configs in a single playwright config file by passing them as projects to defineConfig() and then in package.json specify the project.

```node
const inCluster = {
  name: 'in-cluster',
  timeout: 10 * 1000,
  testDir: 'tests',
  /* Run tests in files in parallel */
  fullyParallel: true,
  retries: process.env.CI ? 2 : 1,
  testMatch: /^(?!.*local-auth|.*reconnect)(.+\.)?(test|spec)\.[jt]s$/,
  use: {
    baseURL: `${protocol}://${host}/`,
  },
}

export default defineConfig({ projects: [main, localAuth, reconnect, inCluster] })
```

```json
"test:integration-in-cluster": "playwright test --project=in-cluster",
```
i think we should go ahead and do that with the this in-cluster config and then in a separate PR add the others.

tasks/setup.yaml
```diff
@@ -26,7 +26,7 @@ tasks:
     description: "start a plain k3d cluster"
     actions:
       - cmd: |
-          k3d cluster delete runtime && k3d cluster create runtime --k3s-arg "--disable=traefik@server:*" --k3s-arg "--disable=servicelb@server:*"
+          k3d cluster delete uds && k3d cluster create uds --k3s-arg "--disable=traefik@server:*" --k3s-arg "--disable=servicelb@server:*"
```
UncleGedd, line 29: Was there a reason to change the name of the cluster from `runtime` to `uds`? The cluster name `uds` conflicts with the slim-dev and core-demo bundles that we deploy

tasks/setup.yaml
```diff
@@ -50,3 +50,9 @@ tasks:
     actions:
       - cmd: uds zarf package create tmp/uds-core/src/authservice --confirm --flavor upstream -o build/
       - cmd: uds zarf package deploy build/zarf-package-uds-core-authservice*.tar.zst --confirm
+
+  - name: core-metrics-server
```
UncleGedd, line 54: This task doesn't seem to work for me, there's no `uds-bundle.yaml` in `hack/test`

tasks/test.yaml
```diff
@@ -75,40 +129,15 @@ tasks:
...
-      - cmd: |
-          helm repo add istio https://istio-release.storage.googleapis.com/charts
-          helm upgrade -i istio-base istio/base -n istio-system --set defaultRevision=default --create-namespace
-        description: install Istio CRDs
-      - cmd: |
-          helm repo add minio https://charts.min.io/
-          helm upgrade -i -n uds-dev-stack minio minio/minio --create-namespace --set replicas=1 --set mode=standalone --set persistence.size=1Gi \
-            --set resources.requests.memory=256Mi --set resources.requests.cpu=100m
-        description: deploy minio to test PVCs
+      - cmd: go test -failfast -v -timeout 30m --tags=integration ./...
```
UncleGedd, line 132: ^ I think this is the result of a merge conflict? We don't have any more build tags

ui/tests/table.spec.ts
```diff
@@ -10,51 +10,57 @@ test.describe('DataTable', async () => {
...
   })
 })
+
+// Checks to see if the table header results are as expected. For example: "Showing 1 of 8"
+async function checkTableHeaderResults(page: Page, expectedActual: number, expectedTotal: number) {
+  const tableHeaderResults = page.getByTestId('table-header-results')
+  await tableHeaderResults.waitFor()
+  const textContent = await tableHeaderResults.textContent()
+
+  console.log('textContent:', textContent)
```
UncleGedd, line 54:
```suggestion
```

tasks/setup.yaml
decleaver: left over code I didn't clean up. Thanks!

tasks/setup.yaml
decleaver: yeah, I updated this because we have some tests that check the node name. This makes it so regardless of how you setup your k3d cluster, all the e2e tests should still pass (at least those tests)

UncleGedd: Does this mean I can't run a slim-dev cluster and run the tests at the same time though?

decleaver: updated tests to handle `runtime` node

ui/playwright.config.incluster.ts
decleaver: So wrestled with this a bit, the webserver config needs to be at the root config level, not the project level. So for the case of in-cluster, we will still need another config file since we don't want to use the webserver config.

UncleGedd: let's go ahead and punt this refactor as part of our refactor Friday

UncleGedd (approved):
🚀

PR #461, opened 2024-10-16
chore: more auth refactors, move Go code into src, remove build tags

UncleGedd:
## Description

- refactors more auth code and adds tests
- moves Go code into `src`
- removes build tags for unit tests and updates tasks


src/pkg/api/resources/cache.go
```diff
@@ -251,11 +251,7 @@ func (c *Cache) bindUDSResources() {
 // setWatchErrorHandler sets a watch error handler on the provided informer for custom resources
 func (c *Cache) setWatchErrorHandler(informer cache.SharedIndexInformer, resource *ResourceList) {
 	err := informer.SetWatchErrorHandler(func(_ *cache.Reflector, _ error) {
-		if !HasCRD(resource.GVR, c.CRDs) {
-			resource.CRDExists = false
-		} else {
-			resource.CRDExists = true
-		}
+		resource.CRDExists = HasCRD(resource.GVR, c.CRDs)
```
TristanHoladay, line 254: :heart_eyes:

ui/src/lib/utils/token-auth.ts
```diff
@@ -1,35 +1,32 @@
...
+    console.error(e)
+    return Promise.reject(e)
   }
 }
 
-const tokenAuth = new TokenAuth()
 const Auth = {
   connect: async (token: string) => {
-    return await tokenAuth.request(token)
+    return await request(token)
```
TristanHoladay, line 29: Should be able to get rid of the `await` here because we await the call to `Auth.connect` in the `layout.ts`.

UncleGedd: good call! I ended removing the `Auth` object entirely (noting that before we weren't getting proper typing on that variable)

TristanHoladay (approved):

Issue #411, opened 2024-10-02
Overview Dashboard: Core Services


DannyDTenacious:
### Describe the solution you'd like

As a consumer of UDS Runtime, I would like to view a summary of core services, status and a corresponding link to view the service (as shown on the Figma design)

### Additional context
https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=4325-5166&m=dev 
<img width="412" alt="Screenshot 2024-10-02 at 2 57 35 PM" src="https://github.com/user-attachments/assets/bc294994-4eb9-4440-9822-2adea8ae7f94">

BillyFigueroa:
This is what the Services view in OpenLens looks like

<img width="1592" alt="Screenshot 2024-10-08 at 1 14 28 PM" src="https://github.com/user-attachments/assets/cb8057ce-c47a-44ce-b199-6681ccd3ea35">

DannyDTenacious:
Per discussion w/Madeline & Gedd in Slack: "we use the names of the functional layers (core services in out github repo); also let's remove the "View Service" link."

BillyFigueroa:
> Per discussion w/Madeline & Gedd in Slack: "we use the names of the functional layers; also let's remove the "View Service" link."
> 
> Reasoning: Apps within functional layers can and will change. we may not always use keycloak for IDAM but does Ashton really need to be burdened by that information?

The links were removed. Can we remove them from the design?

DannyDTenacious:
Madeline has added a note stating the links are for future; eventually this will link to a package definition page with all YAMLs. Does that work?

BillyFigueroa:
> Madeline has added a note stating the links are for future; eventually this will link to a package definition page with all YAMLs. Does that work?

Yes

PR #471, opened 2024-10-18
feat(ui): 411 overview dashboard core services

BillyFigueroa:
## Description

...

## Related Issue

- #411


ui/src/lib/features/k8s/cluster-overview/component.svelte
```diff
@@ -127,6 +135,20 @@
     />
   </div>
 
+  <div class="mt-8 flex flex-col xl:flex-row xl:space-x-4">
+    <div class="w-full xl:w-1/2">
+      <div class="flex p-5 bg-gray-800 rounded-lg overflow-hidden shadow h-full items-center justify-center">
+        Packages Widget Coming Soon...
```
UncleGedd, line 141: This is fine as long as we get the Packages widget out before the next release, otherwise I'd prefer to keep "Coming soon" type language out of the app

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,102 @@
...
+          </div>
+
+          <div class="w-2/12 text-green-400">Running</div>
+        </div>
+      {/each}
+    </div>
+  {/if}
+</div>
+
+<style lang="postcss">
```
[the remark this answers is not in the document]
UncleGedd: Nvm I see it's a style thing, in-line Tailwind classes are fine but this works too

ui/src/lib/features/k8s/cluster-overview/component.svelte
BillyFigueroa: @UncleGedd how do I handle this until then?

1) When is the next release?
2) If it's soon and the packages won't be done, then do I just make this widget full width until the next one is available?

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,102 @@
...
+    'uds-runtime': 'UDS Runtime',
+  }
+
+  const coreServiceKeys = Object.keys(coreServicesMapping)
+
+  let hasNoCoreServices: boolean = false
+  $: hasNoCoreServices = services.every((service) => !coreServiceKeys.includes(service.metadata.name))
+
+  const sortServices = (a: CoreServiceType, b: CoreServiceType) => {
+    if (a.metadata.name < b.metadata.name) {
```
decleaver, line 27: looking for a way to simplify this, found `localeCompare`. Think you can use something like this: `return a.metadata.name.localeCompare(b.metadata.name);`

ui/src/lib/features/k8s/cluster-overview/component.svelte
UncleGedd: 1. Likely this week
2. Full width is great

TristanHoladay:
@BillyFigueroa looks good! do the displayed "names" of the services need to be changed, though, given the Figma design? (e.g. Authorization, Monitoring, Service Mesh...)

UncleGedd:
> @BillyFigueroa looks good! do the displayed "names" of the services need to be changed, though, given the Figma design? (e.g. Authorization, Monitoring, Service Mesh...)

Just chatted with @Madeline-UX about this, thse names should (ideally) line up with the UDS functional layers

BillyFigueroa:
> Also (again just chatted with @Madeline-UX ), let's remove the "View Service >" link for now

The links were removed in this [commit](https://github.com/defenseunicorns/uds-runtime/pull/471/commits/b20a43c7f171ba1d20001cc27ec6db077cd82822). Should we ask design to remove from the mocks?

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,104 @@
...
+  import { type CoreServiceType } from '$lib/types'
+  import { Cube, Information } from 'carbon-icons-svelte'
+
+  export let services: CoreServiceType[] = []
+  export let pods: V1Pod[] = []
+
+  const coreServicesMapping: Record<string, string> = {
+    authservice: 'Authorization',
+    grafana: 'Monitoring',
+    istrio: 'Service Mesh',
```
UncleGedd, line 12: I think you mean `istio`

BillyFigueroa: Good catch. I just updated

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,104 @@
...
+  $: {
+    hasPolicyEngineOperator = pods.filter((pod: V1Pod) => pod?.metadata?.name?.match(/^pepr-uds-core/)).length > 0
+
+    services.forEach((service) => {
+      let serviceName = coreServicesMapping[service.metadata.name]
+
+      uniqueServiceList.push(serviceName)
+      uniqueServiceList = Array.from(new Set([...uniqueServiceList]))
+
+      if (hasPolicyEngineOperator && !uniqueServiceList.includes('Policy Engine & Operator')) {
```
UncleGedd, line 39: I'm guessing this is the edge case for showing Pepr in the Core Services view? Can we add a comment for why this necessary?

ui/src/lib/features/k8s/cluster-overview/component.svelte
```diff
@@ -39,11 +41,26 @@
   let onMessageCount = 0
   let myChart: Chart
   const description = resourceDescriptions['Events']
+  let services: CoreServiceType[] = []
```
UncleGedd, line 44: Can we rename this and related to vars to inclue `core`? That way we don't conflate it K8s Services

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,104 @@
...
+      }
+    })
+  }
+</script>
+
+<div class="core-services">
+  <div class="core-services__header">
+    <h2 class="core-services__header-title">Core Services</h2>
+
+    <Information class="ml-2 w-4 h-4 dark:text-gray-400 text-blue-500" />
```
UncleGedd, line 50: Need to add a description to this tooltip like we have on the other pages, check out `descriptions.ts`

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,104 @@
...
+        <div class="core-services__rows-item">
+          <div class="w-10/12 flex items-center space-x-2">
+            <div class="core-services__name-icon">
+              <Cube size={16} class="text-gray-400" />
+            </div>
+
+            <div class="truncate">{serviceName}</div>
+          </div>
+
+          <div class="w-2/12 text-green-400">Running</div>
```
UncleGedd, line 68: I don't think `Running` should be hardcoded, take a look at the `status.phase` field in the underlying K8s resource

ui/src/lib/components/CoreServicesWidget/component.svelte
BillyFigueroa: This is to avoid duplication. I think I might have a way to make it cleaner. Let me try.

ui/src/lib/features/k8s/cluster-overview/component.svelte
BillyFigueroa: > Can we rename this and related to vars to inclue `core`? That way we don't conflate it K8s Services

updated and pushed up

ui/src/lib/components/CoreServicesWidget/component.svelte
BillyFigueroa: > I don't think `Running` should be hardcoded, take a look at the `status.phase` field in the underlying K8s resource

If, it's not running, is the query going to return the core service? if so, how do we map the "phase" to a color?

Are these considered "Nodes"? so their statuses will be based on this mapping?
Screenshot 2024-10-22 at 10 34 55 AM

UncleGedd: Nope, not Nodes. They are UDS Packages, check out the CRD of a UDS Package (you can do this in the runtime-canary instance)
Screenshot 2024-10-22 at 9 43 07 AM

UncleGedd: > If, it's not running, is the query going to return the core service? if so, how do we map the "phase" to a color?

You should be able to get the `status.phase` of the UDS Package (may need to use the field selectors). For the colors, I'd say:
- Pending: orange
- Ready: green
- Failed red
- Retrying: orange

ui/src/lib/components/CoreServicesWidget/component.svelte
BillyFigueroa: > Need to add a description to this tooltip like we have on the other pages, check out `descriptions.ts`

Asking Mads what the description should be.

Question though, `description.ts` seems to be for actual k8 resources...

this is it's own sort of core service thing. Should this be created as it's own object or do we add ii to `resourceDescription` object?

UncleGedd: Personally I like to keep to all the descriptions in the same place, but if you want to hardcode it in the tooltip for Core Services that's fine too

BillyFigueroa: > Personally I like to keep to all the descriptions in the same place, but if you want to hardcode it in the tooltip for Core Services that's fine too

Added to the current object because it is called resourceDescription and not k8ResourceDescriptions so it's fine to be in there. I guess I could have created a `widgetDescription` object too and we can later

Updated and pushed up

ui/src/lib/components/CoreServicesWidget/component.svelte
BillyFigueroa: > I'm guessing this is the edge case for showing Pepr in the Core Services view? Can we add a comment for why this necessary?

Just made a minor code change to just remove the dupliate rather than check if it already exists in the array and also an explanation for why we are adding `Policy Engine & Operator` to the list if we have pepr pods

TristanHoladay:
@BillyFigueroa sorry to add another comment :sweat_smile: but i think there's 2 issues happening. Apologies if you're already aware of them.

1. looks like the view is not re-rendering when Core services are deploying
Screenshot from 2024-10-22 12-42-15

1. Istio does not have a UDS Package so is not showing in Core Services for Service Mesh. Probably easiest way to get that would be to look for an istiod pod in istio-system namespace / at gateway pods in the istio-admin-gateway or istio-tentant-gateway namespaces.

BillyFigueroa:
> @BillyFigueroa sorry to add another comment 😅 but i think there's 2 issues happening. Apologies if you're already aware of them.
> 
>     1. looks like the view is not re-rendering when Core services are deploying
>        Screenshot from 2024-10-22 12-42-15
> 
>     2. Istio does not have a UDS Package so is not showing in Core Services for Service Mesh. Probably easiest way to get that would be to look for an istiod pod in istio-system namespace / at gateway pods in the istio-admin-gateway or istio-tentant-gateway namespaces.

Hi Tristan,

1) What steps do I take to try to recreate the deploying stage?
2) I ll work on the istio changes

TristanHoladay:
> Hi Tristan,
> 
> 1. What steps do I take to try to recreate the deploying stage?
> 2. I ll work on the istio changes

I did `uds deploy k3d-core-slim-dev:latest --confirm` and then, after the cluster started, deployed Runtime locally.

decleaver:
Not sure if its just me. But pulled down the latest updates, running a newly created k3d cluster, so no core service. But now i cant select any of the views, but if I manually type in the path to view pods for example, then the view menu starts working again.....

BillyFigueroa:
> Not sure if its just me. But pulled down the latest updates, running a newly created k3d cluster, so no core service. But now i cant select any of the views, but if I manually type in the path to view pods for example, then the view menu starts working again.....



> Not sure if its just me. But pulled down the latest updates, running a newly created k3d cluster, so no core service. But now i cant select any of the views, but if I manually type in the path to view pods for example, then the view menu starts working again.....

I could not re-create this. Maybe someone else can try

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,135 @@
...
+    'uds-runtime': 'Frontend Views & Insights',
+  }
+
+  const coreServiceKeys = Object.keys(coreServicesMapping)
+  let hasNoCoreServices: boolean = false
+  let transformedCoreServiceList: { name: string; status: UDSPackageStatus }[] = []
+  let hasPolicyEngineOperator: boolean = false
+  let hasServiceMesh: boolean = false
+
+  $: hasNoCoreServices = coreServices.every((service) => !coreServiceKeys.includes(service.metadata.name))
```
decleaver, line 33:
```suggestion
  $: hasNoCoreServices = Array.isArray(coreServices) && coreServices.every((service) => !coreServiceKeys.includes(service.metadata.name));
```
this along with the suggestion below fixed that issue i was seeing where the view menu's stopped working

ui/src/lib/components/CoreServicesWidget/component.svelte
```diff
@@ -0,0 +1,135 @@
...
+  $: {
+    hasPolicyEngineOperator = pods.filter((pod: V1Pod) => pod?.metadata?.name?.match(/^pepr-uds-core/)).length > 0
+    hasServiceMesh =
+      pods
+        .filter((pod: V1Pod) => pod?.metadata?.name?.match(/^istiod/))
+        .filter((pod) => pod.status?.phase === 'Running').length === 1
+
+    coreServices.forEach((service) => {
```
decleaver, line 41: need to check to make sure you have an array of coreServices. You can wrap this with
`if (Array.isArray(coreServices)) {` to handle potential errors

BillyFigueroa:
@decleaver this might be an ignorant question because I do not understand how to handle errors in go but should this actually be considered an error when we don't have a crd? are most of these resources an array? Could we not just return an empty array in these cases?

I see this is being handled in the k8s/store.ts i.e

```JavaScript
  if ($resources.length > 0 && typeof $resources[0].resource === 'string') {
    if ($resources[0].resource === 'crd not found') {
      this.numResources.set(0)
      return $resources
    }
  }
```

I remember this was causing an issue with one of the tables before.

I will add some logic in the `onmessage` (in cluster-overview) to handle this rather than at the component level, but maybe we should consider a different response? returning error messages to a front end like this is a bit passive. If we can make it an exception or an empty array, I think it would be cleaner

decleaver:
> @decleaver this might be an ignorant question because I do not understand how to handle errors in go but should this actually be considered an error when we don't have a crd? are most of these resources an array? Could we not just return an empty array in these cases?
> 
> I see this is being handled in the k8s/store.ts i.e
> 
> ```js
>   if ($resources.length > 0 && typeof $resources[0].resource === 'string') {
>     if ($resources[0].resource === 'crd not found') {
>       this.numResources.set(0)
>       return $resources
>     }
>   }
> ```
> 
> I remember this was causing an issue with one of the tables before.
> 
> I will add some logic in the `onmessage` (in cluster-overview) to handle this rather than at the component level, but maybe we should consider a different response? returning error messages to a front end like this is a bit passive. If we can make it an exception or an empty array, I think it would be cleaner

Yeah, so this is a little bit of a special case where if we just returned an empty array, we wouldn't be able to tell if the cluster doesn't have any of those resources or if the custom resource definition for that resource exists in the cluster, and if the CRD doesn't exist, we want to let the users know that. This could be something we revisit and find a cleaner way to handle though.

decleaver (approved):
paired to address last fixes

PR #474, opened 2024-10-18
chore(api): bind only to localhost interface on serve

### LINKED ISSUE #440: Don't run the server on all interfaces

We call `http.ListenAndServe(“:8080”)` and do something similar for `ListenAndServeTLS`. The “:8080: syntax serves all host interfaces; lock this down to be only a single interface

TristanHoladay:
## Description
Lock interfaces when serving api to Localhost

## Related Issue

Resolves #440

## Curl Test to private IP In Local Mode
Before: got 200 response with index.html
After: received connection refused

### DIFF
--- a/src/pkg/api/start.go
+++ b/src/pkg/api/start.go
@@ -255,14 +255,14 @@ func fileServer(r chi.Router, root http.FileSystem) error {
 func Serve(r *chi.Mux, localCert []byte, localKey []byte, inCluster bool) error {
 	//nolint:gosec,govet
 	if inCluster {
-		slog.Info("Starting server in in-cluster mode on :8080")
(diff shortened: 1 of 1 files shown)


UncleGedd:
When running in local mode I'm still able to reach the server at `0.0.0.0`

Screenshot 2024-10-18 at 1 18 29 PM

bburky:
@UncleGedd typing https://0.0.0.0 is an unrelated thing (it's usually equivalent to 127.0.0.1 sort of when used like an IP)

What you want to test is that you can't access the website from another computer on your LAN. Or check `netstat -tulnp` and check what the listening IP is (should be 127.0.0.1 now)

UncleGedd (approved):

PR #475, opened 2024-10-18
feat: add user menu and logout

UncleGedd:
## Description

- Adds user menu in top-left of screen for in-cluster deployments with SSO enabled
- Adds logout capability for in-cluster SSO
- Refactors `/auth` endpoint to accomodate local and in-cluster auth (adds lots of testing)
    - Removes the global Session object in favor a of a SessionID string
- Adds SSO to in-cluster tests (including login before tests)

## Related Issue

- https://github.com/defenseunicorns/uds-runtime/issues/489


src/pkg/api/auth/cluster/jwt.go
```diff
@@ -39,16 +47,25 @@ func ValidateJWT(w http.ResponseWriter, r *http.Request) bool {
...
+					// todo: handle multiple groups
+					// set context values on request
+					r = r.WithContext(context.WithValue(r.Context(), groupKey, group))
+					if preferredUsername, ok := token.Claims.(jwt.MapClaims)["preferred_username"].(string); ok {
+						r = r.WithContext(context.WithValue(r.Context(), userKey, preferredUsername))
+					} else {
+						http.Error(w, "Invalid token claims", http.StatusUnauthorized)
+						return r, false
+					}
+					return r, true
```
catsby, line 60: nit: this `return r, true` line executes if line 54 is `ok`, and subsequently line 56 does _not_ execute, so in the interest of simplicity and keeping things close to each other since we're returning anyway, we can reduce a layer of `if {} else {}` like so:

```

```suggestion
					if preferredUsername, ok := token.Claims.(jwt.MapClaims)["preferred_username"].(string); ok {
						r = r.WithContext(context.WithValue(r.Context(), userKey, preferredUsername))
					        return r, true
					} 
					http.Error(w, "Invalid token claims", http.StatusUnauthorized)
					return r, false
```

UncleGedd: updated, appreciate you taking a look!

main.go
```diff
@@ -21,11 +22,11 @@ var localCert []byte
 var localKey []byte
 
 func main() {
-	message.SetLogLevel(message.DebugLevel)
+	slog.SetLogLoggerLevel(slog.LevelDebug)
 
 	r, inCluster, err := api.Setup(&assets)
 	if err != nil {
-		message.WarnErr(err, "failed to start the API server")
+		slog.Warn(fmt.Sprintf("failed to start the API server: %s", err))
```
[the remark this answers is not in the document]
TristanHoladay: oh sorry i didn't look close enough at the previous and see `message`. I thought we were already using slog here, probably cause i did something with slog and these errors in CLI when vendoring.

src/pkg/api/auth/incluster/jwt.go
```diff
@@ -15,13 +16,21 @@ var allowedGroups = []string{
...
+
+const (
+	GroupKey             contextKey = "group"
+	PreferredUserNameKey contextKey = "preferred_username"
+	NameKey              contextKey = "name"
+)
+
 // ValidateJWT checks if the request has a valid JWT token with the required groups.
-func ValidateJWT(w http.ResponseWriter, r *http.Request) bool {
+func ValidateJWT(w http.ResponseWriter, r *http.Request) (*http.Request, bool) {
```
TristanHoladay, line 28: ValidateJWT is getting pretty long, i think breaking it up would be good. Could make that a refactor Friday task if you don't want to do it in this PR.

UncleGedd: Yeah and it's only going to get longer, happy to break it up in a separate PR. I think there's another refactor here to make a common interface out of local vs in-cluster auth to cut down on some of the logic (not in `ValidateJWT` but in the middleware and `/auth` handler)

ui/src/lib/features/navigation/navbar/component.svelte
```diff
@@ -3,10 +3,16 @@
...
   import { authenticated } from '$features/auth/store'
+  import { UserMenu } from '$features/navigation'
+  import type { UserData } from '$features/navigation/types'
   import { NotificationFilled } from 'carbon-icons-svelte'
 
   import { isSidebarExpanded } from '../store'
 
+  export let userData: UserData
+
+  const inClusterAuth = (userData && userData.inClusterAuth) ?? false
```
TristanHoladay, line 14: maybe we don't need this? or perhaps just named a bit different. something about specifying the type of auth as the determining factor feels odd. And though not entirely avoidable, it seems ideal to keep awareness of desktop vs incluster out of the frontend as much as possible.

ui/src/lib/features/navigation/navbar/usermenu/component.svelte
```diff
@@ -0,0 +1,73 @@
...
+  <button
+    on:click={toggleMenu}
+    class="inline-flex bg-gray-800 items-center justify-center w-full rounded-md px-4 py-2 text-sm font-medium text-white focus:outline-none transition-colors duration-200 ease-in-out"
+    class:bg-gray-800={!isOpen}
+    class:hover:bg-gray-700={!isOpen}
+    class:bg-gray-700={isOpen}
+  >
+    <UserAvatarFilled class="h-5 w-5 mr-2" />
+    <span>{userData.name}</span>
+    {#if isOpen}
```
TristanHoladay, line 50: checkout https://github.com/defenseunicorns/uds-runtime/blob/b2efe519b9cd491a4abe83e32ada5db0ede06d63/ui/src/lib/features/navigation/sidebar/component.svelte#L93

double checked with @BillyFigueroa and he agreed that we should probably stick with this pattern.

UncleGedd: fixed 🚀

main.go
catsby, line 29: `slog` enables us to have structured logs, so we should try to use that here. 
```suggestion
		slog.Warn("failed to start the API server", "error", err)
```

Playground example of the differences:

- https://go.dev/play/p/PPRFLglkj1h

`log/slog` docs: 
- https://pkg.go.dev/log/slog#pkg-overview

src/pkg/api/auth/incluster/jwt.go
```diff
@@ -39,16 +48,26 @@ func ValidateJWT(w http.ResponseWriter, r *http.Request) bool {
 		for _, group := range groups {
 			for _, allowedGroup := range allowedGroups {
 				if group == allowedGroup {
-					// Group is allowed
-					return true
+					// Group is allowed, add group and username to request context
+					// set context values on request
+					r = r.WithContext(context.WithValue(r.Context(), GroupKey, group))
+					preferredUsername, preferredUsernameOk := token.Claims.(jwt.MapClaims)["preferred_username"].(string)
```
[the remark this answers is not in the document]
decleaver: only gets run once anyways. disregard

decleaver (dismissed):
👍

src/pkg/api/auth/auth.go
```diff
@@ -0,0 +1,125 @@
...
+// RequestHandler is the main handler for the /auth endpoint; it returns a userResponse struct
+// indicating whether the request was authenticated via local or in-cluster auth, and relevant user data
+func RequestHandler(w http.ResponseWriter, r *http.Request) {
+	resp := UserResponse{
+		InClusterAuth: false,
+	}
+	if config.LocalAuthEnabled && !local.Auth(w, r) {
+		// auth failed, response is already written
+		return
+	} else if config.InClusterAuthEnabled {
```
catsby, line 76: nit/suggestion: the conditional on line 73 doesn't flow into the next statement on 76, so we should break them apart. If line 73 is truthy then we always return on 75, so there's no need for the `else`. 

```suggestion
	if config.LocalAuthEnabled && !local.Auth(w, r) {
		// auth failed, response is already written
		return
	}
	
	if config.InClusterAuthEnabled {
```

There's no hard rule here, but it's mentioned in Effective Go and Google's Go style guide. 

- https://go.dev/doc/effective_go#control-structures
- https://google.github.io/styleguide/go/decisions#indent-error-flow

src/pkg/api/auth/auth.go
```diff
@@ -0,0 +1,125 @@
...
+			http.Error(w, "authorization failure", http.StatusInternalServerError)
+			return
+		}
+	}
+
+	// write response
+	w.WriteHeader(http.StatusOK)
+	bodyBytes, err := json.Marshal(resp)
+	if err != nil {
+		slog.Debug(fmt.Sprintf("failed to marshal response: %s", err.Error()))
```
catsby, line 100:
```suggestion
		slog.Debug("failed to marshal response", "error", err)
```

src/pkg/api/auth/auth.go
```diff
@@ -0,0 +1,125 @@
...
+		localAuthEnabled = true
+	}
+
+	config.LocalAuthEnabled = localAuthEnabled
+	if localAuthEnabled {
+		slog.Info("Local auth enabled")
+		token, err := randomString(96)
+		if err != nil {
+			slog.Error("Failed to generate local auth token")
+			os.Exit(1)
```
catsby, line 48: nit: having `os.Exit(1)` this deep in the code seems odd to me. `main.go` calls `api.Setup()` and checks for errors. The first thing `api.Setup()` does is call `auth.Configure()`, so why not return an error here and let it bubble up to `main`? I would think we should have in `main.go` something like `slog.Info("setting up server..")` and then check for an error.

src/pkg/api/auth/auth_test.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	tests := []struct {
+		name                  string
+		localAuthEnv          string
+		inClusterAuthEnv      string
+		expectedLocalAuth     bool
+		expectedInClusterAuth bool
+		shouldHaveLocalToken  bool
+	}{
```
catsby, line 34: style nit (not particular to this set of tests, but in general): by using a `map[string]struct{}` for test cases, we can reduce the chance of test cases passing as the result of prior tests not cleaning up. Ranging over `map` is _undefined_, meaning the ordering is not consistent, so every the test is ran the ordering could be completely different. You can also conveniently move the `name` attribute to be the `key`, ex:

```go
	tests := map[string]struct {
		localAuthEnv          string
		inClusterAuthEnv      string
		expectedLocalAuth     bool
		expectedInClusterAuth bool
		shouldHaveLocalToken  bool
	}{
		"Default values when env vars not set": {
			localAuthEnv:          "",
			inClusterAuthEnv:      "",
			expectedLocalAuth:     true,
			expectedInClusterAuth: false,
			shouldHaveLocalToken:  true,
		},
	},
       [....]
```

src/pkg/api/auth/local/local.go
```diff
@@ -0,0 +1,69 @@
...
+	token := r.URL.Query().Get("token")
+	if token == "" {
+		// Handle session cookie validation
+		if valid := ValidateSessionCookie(w, r); valid {
+			w.WriteHeader(http.StatusOK)
+			return true
+		}
+		w.WriteHeader(http.StatusUnauthorized)
+		return false
+	} else if token != AuthToken {
```
catsby, line 31: Same comment about the `if { return } else { return }`, we should break them appart

src/pkg/api/auth/auth.go
UncleGedd: Good call, updated

src/pkg/api/auth/local/local.go
UncleGedd: Good catch, updated

src/pkg/api/auth/auth_test.go
UncleGedd: Gotcha, so the idea is that by introducing randomness, we can reduce false positives. I'll make an issue for this one to address this across the code base

src/pkg/api/auth/local/local.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		w.WriteHeader(http.StatusUnauthorized)
+		return false
+	}
+	if token != AuthToken {
+		w.WriteHeader(http.StatusUnauthorized)
+		return false
+	}
+	// valid token, generate session id and set cookie
+	once.Do(func() {
```
catsby, line 37: What is the intent of the `once.Do()` here? If I read this correctly, we should only get to this point if `token == AuthToken` one time anyway, otherwise it will be validated on line 25 or otherwise rejected.

catsby (commented):
Not totally done reviewing but wanted to ask this before I continued

src/pkg/api/auth/local/local.go
catsby: I wrote something up but it was too long so lets sync sometime. The summary is that the `sync.Once` here is safe to remove, it isn't really doing anything for us because if they already have a session id, they wont end up here. Happy to go in to detail though.

UncleGedd: I seen what you mean, the logic will only let it trigger once anyway, I'll remove it

ui/src/routes/auth.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      name: '',
+      'preferred-username': '',
+      group: '',
+      'in-cluster-auth': false,
+    }
+
+    // Mock successful fetch response
+    fetchMock.mockResolvedValueOnce({
+      ok: true,
+      json: () => Promise.resolve(mockUserData),
```
TristanHoladay, line 59: nit more FYI than request to change: looks like you don't need Promise.resolve() for this.
tests still pass as
```suggestion
      json: () => mockUserData,
```

ui/src/routes/auth.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    consoleSpy.mockRestore()
+  })
+
+  test('malformed JSON in successful response', async () => {
+    mockUrl.searchParams.set('token', 'valid-token')
+    Object.defineProperty(window, 'location', {
+      value: { href: mockUrl.href },
+      writable: true,
+    })
```
TristanHoladay, line 193:
```suggestion
```

ui/src/routes/auth.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      },
+    })
+  })
+
+  test('authentication failure with invalid token', async () => {
+    mockUrl.searchParams.set('token', 'invalid-token')
+    Object.defineProperty(window, 'location', {
+      value: { href: mockUrl.href },
+      writable: true,
+    })
```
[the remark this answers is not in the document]
UncleGedd: done!

ui/src/routes/auth.test.ts
UncleGedd: done! removed here and in other tests, good catch

TristanHoladay (approved):

Issue #432, opened 2024-10-11
Remove "message" package


UncleGedd:
### Describe what should be investigated or refactored

Consolidate all UDS Runtime server logging into `slog` and remove the `message` package entirely

PR #498, opened 2024-10-24
feat(api): handle local mode in airgap

UncleGedd:
## Description

- Ensures we can run the app locally in the airgap
- Refactors config logic into an `init` function


src/pkg/api/airgap.go
```diff
@@ -0,0 +1,140 @@
...
+			os.Exit(1)
+		}
+	}()
+	<-stop
+	slog.Info("Shutting down server...")
+	return nil
+}
+
+// isAirgapped checks if we're in the airgap by checking Google and Cloudflare DNS servers
+func isAirgapped(timeout time.Duration) bool {
```
UncleGedd, line 48: TODO: dep inject the IPs to check so we can test properly

UncleGedd: also this function is gross 😅

UncleGedd: nvm, simplified this fn to the point it's not worth testing

UncleGedd:
Chatted with @TristanHoladay , going to make the cert + domain configurable, this will:
- make testing easier
- allow customer configuration in `uds ui` to avoid the HTTPS errors when in the airgap

src/pkg/api/airgap.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	defer cleanupCerts()
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		//nolint:gosec,govet
+		err := http.ListenAndServeTLS(":8443", "airgap-cert.pem", "airgap-key.pem", r)
+		if err != nil {
+			slog.Error("Failed to start server:", "error", err)
+			os.Exit(1)
```
catsby, line 39: If we hit this `os.Exit(1)`, the deferred `cleanupCerts()` func will not run

src/pkg/api/start.go
```diff
@@ -258,36 +258,40 @@ func fileServer(r chi.Router, root http.FileSystem) error {
...
-		// Create a server with TLS config
-		server := &http.Server{
-			Addr:      "127.0.0.1:8443",
-			Handler:   r,
-			TLSConfig: tlsConfig,
-		}
+	// connected to internet, create tls config from embedded cert and key
+	cert, err := tls.X509KeyPair(localCert, localKey)
+	if err != nil {
+		log.Fatalf("Failed to load embedded certificate: %v", err)
```
catsby, line 279: `log.Fatalf()` logs a message then invokes `os.Exit(1)`. We invoke `api.Serve()` from `main.go` ~line 33 where, where if `Serve()` returns an error we `os.Exit(1)` so we can simply log the error here and return it. This keeps up with using `slog` while passing control back to `main.go` 

```suggestion
		slog.Error("failed to load embedded certificate", "error", err)
		return err
```

src/pkg/api/airgap.go
UncleGedd: fixed!

src/pkg/api/start.go
```diff
@@ -258,36 +258,40 @@ func fileServer(r chi.Router, root http.FileSystem) error {
...
+		TLSConfig:         tlsConfig,
+		ReadHeaderTimeout: 10 * time.Second,
+	}
 
-		if err = server.ListenAndServeTLS("", ""); err != nil {
-			message.WarnErrf(err, "server failed to start: %s", err.Error())
-			return err
-		}
+	if err = server.ListenAndServeTLS("", ""); err != nil {
+		message.WarnErrf(err, "server failed to start: %s", err.Error())
```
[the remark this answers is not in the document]
UncleGedd: > If we're just watching the logs via another means (can we?), will this message show?

Today I don't think we handle the case of watching logs from another means. Noting that we have an issue (#432) to remove use of `message` entirely

decleaver (approved):
👍

catsby (approved):

PR #499, opened 2024-10-24
feat: switch currently viewed cluster

### LINKED ISSUE #348: Enable Switching between Clusters

As an SRE, I find it frustrating and time-consuming to constantly switch between different interfaces when I need to manage multiple Kubernetes clusters. Currently, Runtime doesn't provide a way to easily switch between clusters. 

- **Given** the SRE is logged into Runtime and viewing information for a specific Kubernetes cluster
- **When** they select a different cluster from a dropdown menu or similar UI element
- **Then** Runtime should
  - Switch the context to the selected cluster
  - Update all relevant information and metrics to reflect the newly selected cluster
  - Maintain any applied filters or views that are cluster-agnostic
  - **Provide a clear visual indication of which cluster is currently active**

- Creating separate dashboard instances for each cluster: This would be less efficient and could lead to increased resource usage.
- Using browser tabs to manage different clusters: This approach lacks integration and doesn't provide a seamless experience. Users may decide to do this anyway and may be worth exploring how it is being used in prod.

### Additional context

- The feature should support switching between all clusters that the user has permission to access.
- Consider adding a "favorites" or "recent clusters" list for quick access to frequently used clusters. (validation needed, not in scope)
- Ensure that the cluster switching process is fast and doesn't disrupt the user's workflow.

TristanHoladay:
## Description
Adds the ability for users to change which cluster they are viewing in Runtime by selecting from a list produced from the available clusters in their local Kubeconfig.

See [Design Doc](https://docs.google.com/document/d/1YUW1rAsD_t4gqFBVMVO9GPjN7KjZh7LabAILTMkKUjQ/edit?tab=t.0#heading=h.c80zcnxqkbq1) for more details.

## Related Issue

Resolves #348 

Switch -- Error state then Success
[Screencast from 11-05-2024 10:35:53 AM.webm](https://github.com/user-attachments/assets/2c8d4e1e-7733-4cf3-a2a6-f083ea680258)

Switch after disconnection event occurred:
[Screencast from 11-05-2024 10:37:07 AM.webm](https://github.com/user-attachments/assets/13f59cbf-7e01-4987-8504-bb815e7bc720)


src/pkg/api/docs/docs.go
```diff
@@ -28,6 +28,52 @@ const docTemplate = `{
                 }
             }
         },
+        "/api/v1/cluster": {
+            "post": {
+                "description": "Swith the currently viewed cluster",
```
catsby, line 33:
```suggestion
                "description": "Switch the currently viewed cluster",
```

src/pkg/api/docs/swagger.json
```diff
@@ -17,6 +17,52 @@
                 }
             }
         },
+        "/api/v1/cluster": {
+            "post": {
+                "description": "Swith the currently viewed cluster",
```
catsby, line 22:
```suggestion
                "description": "Switch the currently viewed cluster",
```

src/pkg/api/docs/swagger.yaml
```diff
@@ -10,6 +20,35 @@ paths:
           description: OK
       tags:
       - auth
+  /api/v1/cluster:
+    post:
+      consumes:
+      - application/json
+      description: Swith the currently viewed cluster
```
catsby, line 27:
```suggestion
      description: Switch the currently viewed cluster
```

src/pkg/api/handlers.go
```diff
@@ -935,3 +936,22 @@ func healthz(w http.ResponseWriter, _ *http.Request) {
...
+// @Description get list of clusters with context names from local kubeconfig
+// @Tags clusters
+// @Produce json
+// @Success 200
+// @Router /api/v1/clusters [get]
+func getClusters(ks *session.K8sSession) func(w http.ResponseWriter, r *http.Request) {
+	return client.ServeClusters(ks.InCluster, ks.CurrentCtx)
+}
+
+// @Description Swith the currently viewed cluster
```
catsby, line 949:
```suggestion
// @Description Switch the currently viewed cluster
```

I think this is the only one you need to actually accept, the others are generated?

catsby: Update: I deleted my other suggestions as this is likely the only one needed

src/pkg/k8s/client/client.go
```diff
@@ -4,11 +4,15 @@
...
 import (
+	"encoding/json"
 	"fmt"
+	"log/slog"
+	"net/http"
 
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
```
catsby, line 15: nit: us mixed caps e.g. `clientCmdAPI`

src/pkg/k8s/client/client.go
```diff
@@ -74,3 +82,65 @@ var GetCurrentContext = func() (string, string, error) {
...
+		}
+
+		contexts, err := Contexts()
+		if err != nil {
+			slog.Error("Failed to get contexts", "error", err)
+			http.Error(w, "Failed to get contexts", http.StatusInternalServerError)
+			return
+		}
+
+		slog.Debug("ServeClusters called", "inCluster", inCluster, "currentContext", currentContext)
```
catsby, line 113: nit: move this to the top of the function that's returned, line ~88, so that we always log the debug message that we came here, otherwise if we hit an error or if `inCluster` is true, we'll likely not hit this message

src/pkg/k8s/client/client.go
```diff
@@ -74,3 +82,65 @@ var GetCurrentContext = func() (string, string, error) {
...
+			if err != nil {
+				slog.Error("Failed to marshal empty data", "error", err)
+				http.Error(w, "JSON marshalling error", http.StatusInternalServerError)
+			}
+
+			// Write the data to the response
+			if _, err := w.Write(data); err != nil {
+				http.Error(w, "Failed to write data", http.StatusInternalServerError)
+				return
+			}
```
catsby, line 103: Should we `return` here? The code comment suggests if we're `inCluster` that we should just return an empty slice and move on, but we don't actually `return`

src/pkg/k8s/client/client.go
```diff
@@ -74,3 +82,65 @@ var GetCurrentContext = func() (string, string, error) {
...
+func IsRunningInCluster() (bool, error) {
+	_, err := rest.InClusterConfig()
+
+	if err == rest.ErrNotInCluster {
+		return false, nil
+	} else if err != nil {
+		return true, err
+	}
+
+	return true, nil
```
catsby, line 145: nit: we can simplify this with 1 fewer logic check and return statement 
```suggestion
	if err == rest.ErrNotInCluster {
		return false, nil
	} 

	return true, err
```

src/pkg/k8s/session/session.go
```diff
@@ -212,3 +218,80 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
...
+			http.Error(w, "failed to read request body", http.StatusBadRequest)
+			return
+		}
+		defer r.Body.Close()
+
+		var reqData map[string]client.ClusterInfo
+		if err := json.Unmarshal(body, &reqData); err != nil {
+			http.Error(w, "invalid request format", http.StatusBadRequest)
+			return
+		}
```
catsby, line 240: nit: I think we can simplify this like so: 

```suggestion
		defer r.Body.Close()

		var reqData map[string]client.ClusterInfo
		if err := json.NewDecoder(r.Body).Decode(&reqData); err != nil {
			http.Error(w, "invalid request format", http.StatusBadRequest)
			return
		}
```

src/pkg/k8s/session/session.go
```diff
@@ -212,3 +218,80 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
...
+
+		slog.Info("Switching cluster", "clusterInfo", reqData["cluster"])
+
+		contexts, err := client.Contexts()
+		if err != nil {
+			http.Error(w, "failed to get contexts", http.StatusInternalServerError)
+			return
+		}
+
+		targetContext, targetCTXName := findIn(contexts, reqData)
```
catsby, line 250: nit: because the `findIn` function simply ranges over the contents of a map (here `contexts`), we should just use the "comma, ok" idiom:

- https://go.dev/doc/effective_go#maps

ex:

```go
// should targetCtxName be targetClusterName?
var targetCtx, targetCtxName string
if targetCtx, ok := contexts[reqData["cluster"].Context]; ok {
  // verify the target cluster name
  if targetCtx.Cluster == reqData["cluster"].Name {
    targetCtxName = reqData["cluster"].Context
  }
}
```

Maps are Golang's hashtables and they're very good/fast when searching/indexing by key, as opposed to ranging over them and trying every combination. Because iterating over a map has an "undefined" ordering, every time you range over them you could get a completely different order, so searching for a key by ranging will give you inconsistent performance. 

Note that as written `findIn()` doesn't account for the situation where the context name is found, but the cluster names don't match.

src/pkg/k8s/session/session.go
```diff
@@ -212,3 +218,80 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
...
+
+		// update the K8sSession
+		ks.Clients = targetClients
+		ks.Cache = targetCache
+		ks.CurrentCtx = targetCTXName
+		ks.CurrentCluster = targetContext.Cluster
+
+		slog.Info("Successfully switched cluster", "context", targetCTXName, "cluster", targetContext.Cluster)
+
+		w.WriteHeader(http.StatusOK)
```
catsby, line 277: nit: this *shouldn't* be necessary; the write on ~278 automatically sets the header to `StatusOK` if `WriteHeader()` has not already been called. 
```suggestion
```

src/test/e2e/api_test.go
```diff
@@ -344,6 +344,32 @@ func TestClusterHealth(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+
+		rr := httptest.NewRecorder()
+		req := httptest.NewRequest("GET", "/api/v1/clusters", nil)
+
+		// Start serving the request for 1 second
+		go func(ctx context.Context) {
+			r.ServeHTTP(rr, req)
+		}(ctx)
```
catsby, line 364: We pass in `ctx` here but it's not actually used, unless I misunderstand something here

src/pkg/k8s/client/client.go
TristanHoladay: oooo good catch

src/pkg/k8s/client/client.go
TristanHoladay: nice, i like that

src/test/e2e/api_test.go
```diff
@@ -344,6 +344,32 @@ func TestClusterHealth(t *testing.T) {
 	})
 }
 
+func TestGetClusters(t *testing.T) {
+	r, err := setup()
+	require.NoError(t, err)
+
+	defer teardown()
```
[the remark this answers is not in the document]
TristanHoladay: hmm, i think `teardown()` was removed in a recent merge.

src/pkg/k8s/session/session.go
TristanHoladay: gotcha. i'd read a little before about the "undefined" ordering but i didn't think about how that can affect performance on each run. 

Looking at it now, i think i was originally looking up from the client.Contexts() because i wasn't adding context name to ClusterInfo and realized i need the name for passing to the overrides for the new client. Now that the context is part of the request body, we can actually get rid of this look up entirely i think. If the context no longer exists or the data is wrong for some reason the error check for creating the client should catch it and tell the user.

src/test/e2e/api_test.go
TristanHoladay: yeah i've been basing these tests off of previous SSE tests that showed passing in the ctx. removing it i see that the test still passes as long as ctx.Done() is still called. That's interesting. I'm going to make an issue for us to take a look at this again since it's in all of our api tests.

src/pkg/api/start.go
```diff
@@ -304,3 +306,9 @@ func withLatestCache(k8sSession *session.K8sSession, handler func(cache *resourc
 		handler(k8sSession.Cache)(w, r)
 	}
 }
+
+func withLatestSession(k8sSession *session.K8sSession, handler func(k8sSession *session.K8sSession) func(w http.ResponseWriter, r *http.Request)) http.HandlerFunc {
```
TristanHoladay, line 310: @catsby this is the wrapper. because the handlers return functions that return a `http.HandlerFunc` some kind of closure is happening that evaluates arguments to the handler only once. So the reference to the `K8sSession.Cache`, for example, never updates even if we have indeed switched out the cache for a new one. By using a wrapper like `withLatestSession` or `withLatestCache` that sets up the handler as a lambda (iirc) then we can pass in an argument like `K8sSession` and it's reference is updated every time the handler runs. 

It looks like we can get around this when the function called by the handler is a receiver function. For instance, `session.Switch()` does not need to be wrapped in the lambda and we are still able to access the latest version of `K8sSession`. Is that worth leaning toward receiver functions for cases like this, or should we prefer not to use receiver functions and accept that we need the lambda wrapper?

UncleGedd:
cc'ing @Madeline-UX for a design review, I think at minimum we can make that dropdown look cleaner, maybe make the animation have a transparent background, and clean up the icon usage

UncleGedd:
Really appreciate the videos by the way!

BillyFigueroa:
@TristanHoladay are you handling the case where a cluster is displayed on the menu but the cluster is stopped? For example, I was just running uds-runtime cluster and have the uds cluster but is stopped. The stopped cluster still shows up on the drop down menu and when I click it just goes on forever. Video included

https://github.com/user-attachments/assets/d5a3f032-3520-4655-acb2-7dcbbf940801

BillyFigueroa:
This is a question...

Do we need the name to display twice like it does now? I think the prefix is the context, so will that ever be different than the second half which I guess is the cluster name?

i.e.
instead of `k3d-uds.k3d-uds` it's just `k3d-uds`

Screenshot 2024-11-04 at 12 48 50 PM

TristanHoladay:
> @TristanHoladay are you handling the case where a cluster is displayed on the menu but the cluster is stopped? For example, I was just running uds-runtime cluster and have the uds cluster but is stopped. The stopped cluster still shows up on the drop down menu and when I click it just goes on forever. Video included

nice catch! i know exactly what this is cause the same thing happened when i had to update the reconnection logic. We're able to successfully make the client because it exists in the config but when the cache gets made it sits indefinitely in the `cache.WaitForCacheSync()` because none of the informers can connect to the cluster. i'll update with the fix.

TristanHoladay:
> @TristanHoladay are you handling the case where a cluster is displayed on the menu but the cluster is stopped? For example, I was just running uds-runtime cluster and have the uds cluster but is stopped. The stopped cluster still shows up on the drop down menu and when I click it just goes on forever. Video included

Avoids starting new cache if connection to client is refused.
Screenshot from 2024-11-04 13-53-12

TristanHoladay:
> This is a question...
> 
> Do we need the name to display twice like it does now? I think the prefix is the context, so will that ever be different than the second half which I guess is the cluster name?
> 
> i.e. instead of `k3d-uds.k3d-uds` it's just `k3d-uds`
> 
> Also, do we want to make the font inside the drop down a little smaller? the text on the dropdown seems smaller

It is possible to have a different context name than cluster name, and its imperative we have that for the backend. As far as showing that to the users :man_shrugging: It is possible that you have 2 different contexts each with a cluster named the same. So at least for duplicate cluster names the context name will be important. @Madeline-UX do we want to show context name only if there are other contexts with duplicate cluster names?

Also, for the font, Figma shows them as the same font size but i don't have a strong opinion there.

decleaver:
Unable to reconnect to original cluster after attempting to connect to bad cluster. 
Steps to reproduce:
1. Start runtime connected to a good cluster
2. Attempt to switch to bad cluster

You get an error message, the selected cluster in the dropdown remains as the original good cluster but there is no way to get it to reconnect, without refreshing the browser

Would recommend the drop down updating to the bad cluster you tried to connect to, that way you can select the working cluster to reconnect.

image

TristanHoladay:
> Unable to reconnect to original cluster after attempting to connect to bad cluster. Steps to reproduce:
> 
> 1. Start runtime connected to a good cluster
> 2. Attempt to switch to bad cluster
> 
> You get an error message, the selected cluster in the dropdown remains as the original good cluster but there is no way to get it to reconnect, without refreshing the browser
> 
> Would recommend the drop down updating to the bad cluster you tried to connect to, that way you can select the working cluster to reconnect.
> 
> image

oh man good catch! i think the issue is not that the connection to the original cluster doesn't work (the K8sSession hasn't been updated if the switch fails), but that there's no mechanism for setting the loading to false in this case, so it remains on the error version of the loading page indefinitely. 

Changing the selected cluster does make sense except we'd have to update the K8sSession with no client or cache just to re-create the good one again. I'm going to see if there's a different solution first.

decleaver:
can make this a follow on issue, but i have old contexts from CMS days that are really long. Looks like the dropdown truncates these (which looks nice but doesn't let me see the whole name). Maybe add a tool tip on hover to see entire name? The selected context shows the full name which looks kind of weird, so not sure how we want to handle it.

image

TristanHoladay:
> can make this a follow on issue, but i have old contexts from CMS days that are really long. Looks like the dropdown truncates these (which looks nice but doesn't let me see the whole name). Maybe add a tool tip on hover to see entire name? The selected context shows the full name which looks kind of weird, so not sure how we want to handle it.
> 

yeah i think there are some design decisions i want to get with @Madeline-UX on. i'll make a separate issue and add this to it. Thanks for checking that though! super helpful to see more real-life cluster names.

src/pkg/k8s/client/client.go
```diff
@@ -74,3 +87,63 @@ var GetCurrentContext = func() (string, string, error) {
...
+		slog.Debug("ServeClusters called", "inCluster", inCluster, "currentContext", currentContext)
+		w.Header().Set("Content-Type", "text/json; charset=utf-8")
+		w.Header().Set("Cache-Control", "no-cache")
+
+		// running in cluster return []
+		if inCluster {
+			data, err := json.Marshal([]interface{}{})
+			if err != nil {
+				slog.Error("Failed to marshal empty data", "error", err)
+				http.Error(w, "JSON marshalling error", http.StatusInternalServerError)
```
decleaver, line 102: Should we return here if there is an error, so it doesn't try to write potentially bad data down below?

src/pkg/k8s/session/session.go
```diff
@@ -30,14 +32,14 @@ type K8sSession struct {
 	createClient   createClient
 }
 
-type createClient func() (*client.Clients, error)
+type createClient func(overrides *clientcmd.ConfigOverrides) (*client.Clients, error)
 type createCache func(ctx context.Context, client *client.Clients) (*resources.Cache, error)
 
 var lastStatus string
```
decleaver, line 38: Also can be another issue since not a part of this PR, but I think lastStatus can run into race conditions with StartClusterMonitoring running on a ticker and the infinite loop in HandleReconnection()

TristanHoladay: i'll look into it again, but i think things should be good because we block cluster monitoring for both switch and reconnection and then we break out of the reconnection if status is switched.

src/pkg/k8s/session/session.go
```diff
@@ -212,3 +216,73 @@ func (ks *K8sSession) ServeConnStatus() http.HandlerFunc {
...
+
+		if ks.InCluster {
+			http.Error(w, "cannot switch clusters while in-cluster", http.StatusBadRequest)
+			return
+		}
+
+		// Set ready to false to block cluster check ticker since switching
+		ks.ready = false
+
+		defer r.Body.Close()
```
catsby, line 234: nit: this _should_ be safe to remove. 
```suggestion
```
See https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/net/http/request.go;l=179-182

TristanHoladay: Thanks @catsby. i'll add this to the follow on PR i'll be doing.

Issue #113, opened 2024-07-31
View Logs in Tab view


adam-defenseunicorns:
### Is your feature request related to a problem? Please describe

I'm always frustrated when I am unable to have easy access to the logs in order to troubleshoot effectively. 

### Describe the solution you'd like

- **Given** a detail view in Runtime
- **When** I click on a log tab
- **Then** I see the relevant log output in a sidebar

### Describe alternatives you've considered

(optional) A clear and concise description of any alternative solutions or features you've considered.

adam-defenseunicorns:
@Madeline-UX is this something you were planning on working on over the next few days or does this already exist?

adam-defenseunicorns:
@DannyDTenacious can you please update this issue with progress and any details?

DannyDTenacious:
Figma design: **https://www.figma.com/design/zmKcJ9Xin7ChzyGy6RCFLe/UDS-Runtime-(UI%2FCLI)?node-id=5256-13586&m=dev**

Met w/Gedd, Tristan & Adam (10/15) to discuss solution options; decided on triggering "Open Logs" in the dense view drawer, opening a new window for that specific log. Documentation is also in the Figma.

PR #506, opened 2024-10-29
feat(ui): logs view

UncleGedd:
## Description

Logs view with the following features:
- Auto scroll
- Search
- Download
- Copy

Screenshot 2024-11-04 at 8 23 48 AM

## Related Issue

- https://github.com/defenseunicorns/uds-runtime/issues/113


src/pkg/api/rest/filter.go
```diff
@@ -28,6 +28,7 @@ func jsonMarshal(payload any, fieldsList []string) ([]byte, error) {
 		// Handle multiple resources
 		case []unstructured.Unstructured:
 			filteredItems := filterItemsByFields(payload, fieldsList)
+			// todo: this line fails sometimes with: fatal error: concurrent map iteration and map write
```
catsby, line 31: Is there a trick to triggering this error? I tried running the test with high counts but haven't seen it yet. If you know a way to cause this I'd be happy to dig-in

src/pkg/api/handlers.go
```diff
@@ -136,6 +136,39 @@ func getPod(cache *resources.Cache) func(w http.ResponseWriter, r *http.Request)
...
+		logs := []string{
+			"2024-01-01T00:00:00Z [INFO] This is a log message",
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		err := json.NewEncoder(w).Encode(logs)
+		if err != nil {
+			return
+		}
```
catsby, line 168: If there's an error we should set the header status. 

```suggestion
		err := json.NewEncoder(w).Encode(logs)
		if err != nil {
                        slog.Error("failed to encode getPodLog response", "error", err)
		        http.Error(w, "Internal Server Error", http.StatusInternalServerError)
			return
		}
```
Note that if it's successful we don't need to send `StatusOK`, it's implied on the first write:

>  If WriteHeader is not called explicitly, the first call to Write will trigger an implicit WriteHeader(http.StatusOK).

- https://pkg.go.dev/net/http#pkg-types

src/pkg/api/rest/filter.go
UncleGedd: It seems pretty random....I just let the app run for a while and it eventually crashes (noting that I'm actively doing frontend dev when this happens)

src/pkg/api/logs/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	reader := bufio.NewReader(logStream)
+	flusher := w.(http.Flusher)
+
+	// Start log reading goroutine
+	go func() {
+		for {
+			line, err := reader.ReadBytes('\n')
+			if err != nil {
+				if err != io.EOF && !errors.Is(err, context.Canceled) {
+					buffer.WriteString(fmt.Sprintf("ERROR: %v\n", err))
```
catsby, line 72: nit: use `Fprintf`
```suggestion
					fmt.Fprintf(buffer, "ERROR: %v\n", err)
```

src/pkg/api/logs/logs.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+				close(shouldFlush)
+				return
+			}
+
+			buffer.Write(line)
```
catsby, line 78: There's a data race between here and line 98; in 1 goroutine we're writing and in another we're (highly likely) reading at the same time from a buffer that's shared across the 2 via enclosure

catsby (commented):
The `handleLogStream()` function has a data race in it I think we need to sort out

src/pkg/api/logs/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	flusher := w.(http.Flusher)
+
+	// Start log reading goroutine
+	go func() {
+		for {
+			line, err := reader.ReadBytes('\n')
+			if err != nil {
+				if err != io.EOF && !errors.Is(err, context.Canceled) {
+					buffer.WriteString(fmt.Sprintf("ERROR: %v\n", err))
+				}
```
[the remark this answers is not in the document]
UncleGedd: > Is there a reason for the 2 different approaches?

I don't think so, I probably just forgot to standardize

src/pkg/api/logs/logs.go
```diff
@@ -0,0 +1,152 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// Set up streaming for new logs
+	podLogs := clientSet.CoreV1().Pods(namespace).GetLogs(pod, &v1.PodLogOptions{
+		Container: container,
+		Follow:    true,
+		SinceTime: &metav1.Time{Time: lastReadTime},
+	})
```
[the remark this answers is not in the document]
UncleGedd: I think this is because you need to change the `PodLogOptions` each time. For the existing, logs we don't set `Follow: true` but for new logs we do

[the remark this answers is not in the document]
catsby: Can we combine all 3 options? It seems to reason we could just make 1 stream request that we `Follow`; give us the last `100` lines from `Since 1 hour`?

src/pkg/api/logs/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	shouldFlush := make(chan struct{}, 1)
+	reader := bufio.NewReader(logStream)
+	flusher := w.(http.Flusher)
+
+	// Start log reading goroutine
+	go func() {
+		for {
+			line, err := reader.ReadBytes('\n')
+			if err != nil {
+				if err != io.EOF && !errors.Is(err, context.Canceled) {
```
catsby, line 71: Is `contextCanceled` returned from the `reader` when `logStream`'s context is canceled?

UncleGedd: Good catch, I think the `contextCanceled` is a holdover `ioutils`. I'll remove that check for the `context.Canceled`

src/pkg/api/logs/logs.go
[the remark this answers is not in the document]
UncleGedd: Looking at this again, for `streamExistingLogs`, we grab all the existing logs, so no need to delimit by a newline; this enables us to send all the existing logs in a single SSE event. For streaming new logs, we send one log per SSE event, where each log is delimited by a newline

src/pkg/api/logs/logs.go
UncleGedd: fixed with latest commit, the goroutine produces logs and sends them to channel, while the main for/select loop reads from the channel; in this version, only the for/select loop touches the buffer

ui/playwright.config.ts
```diff
@@ -27,7 +27,7 @@ const defaultConfig: PlaywrightTestConfig = {
   testDir: 'tests',
   /* Run tests in files in parallel */
   fullyParallel: true,
-  retries: process.env.CI ? 2 : 1,
+  retries: process.env.CI ? 3 : 1,
```
BillyFigueroa, line 30: Can you explain why are we adding another retry here?

UncleGedd: The e2e logs test is flaky when loading the Pepr logs, it passes consistently on my machine but in CI I saw it fail once. Increasing the retries was my attempt to mitigate SSE flakiness in tests. Open to other suggestions though!

ui/src/lib/components/k8s/Drawer/component.svelte
```diff
@@ -120,7 +120,7 @@
       <div class="flex font-medium pt-3">
         <ul class="flex w-full" id="drawer-tabs">
           <li class="flex-1">
-            <button id="metadata" class:active={activeTab === 'metadata'} on:click={setActiveTab}>Metadata</button>
+            <button id="details" class:active={activeTab === 'details'} on:click={setActiveTab}>Details</button>
```
BillyFigueroa, line 123: Was the decision to switch to Details discussed with the team?

UncleGedd: Yes, since we put the "View Logs" button in that tab, it's no longer strictly "Metadata"

src/pkg/api/logs/logs.go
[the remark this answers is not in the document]
UncleGedd: discussed sync

src/pkg/api/logs/logs.go
UncleGedd: refactored again

ui/src/lib/components/k8s/Logs/component.svelte
```diff
@@ -0,0 +1,277 @@
...
+      // Hide notification after 2 seconds
+      notificationTimeout = setTimeout(() => {
+        showNotification = false
+      }, 2000)
+    } catch (err) {
+      console.error('Failed to copy logs:', err)
+    }
+  }
+
+  export const handleDownload = () => {
```
BillyFigueroa, line 59: Could we move this function to the `logs.ts` ?

ui/src/lib/components/k8s/Logs/component.svelte
```diff
@@ -0,0 +1,277 @@
...
+  } from '$components/k8s/Logs/store'
+  import type { Pod } from '$lib/types'
+  import { CheckmarkOutline, Copy, Download, Search, Terminal } from 'carbon-icons-svelte'
+
+  // manage notifications
+  let showNotification = false
+  let notificationTimeout: ReturnType<typeof setTimeout>
+  let notificationMessage = ''
+
+  const handleCopy = async () => {
```
BillyFigueroa, line 38: Could we move this function to the `logs.ts` ?

ui/src/lib/components/k8s/Logs/component.svelte
```diff
@@ -0,0 +1,277 @@
...
+      }
+      window.removeEventListener('keydown', handleKeydown)
+    }
+  })
+</script>
+
+<section class="h-[35%] flex-shrink-0 mt-6" data-testid="logs-view">
+  <div class="flex h-full w-full flex-col bg-gray-800 rounded-lg overflow-y-auto">
+    <!-- Header -->
+    <div class="flex items-center justify-between py-2 border-b border-gray-700 mx-4">
```
BillyFigueroa, line 171: I think this needs a little more vertical padding. `y-3` for example. Look at the header for the table. it has the same vertical space. We should be consistent with padding as much as possible

ui/src/lib/components/k8s/Logs/logs.ts
```diff
@@ -0,0 +1,129 @@
+// Copyright 2024 Defense Unicorns
```
BillyFigueroa, line 1: What do you think about naming this `helpers.ts` ?

UncleGedd: No preference, `helpers.ts` is fine, I was just aggregating all the non-store functions/handlers for this view

ui/src/lib/components/k8s/Logs/component.svelte
UncleGedd: We can't without a bit of pain. Because those notifications have timeouts that we want to clean up when the page exits, they are kind of a pain to move around. One option is to make them stores but it felt heavy handed

ui/src/lib/components/k8s/Logs/component.svelte
UncleGedd: same comment as below

ui/src/lib/components/k8s/Logs/component.svelte
UncleGedd: good call, fixed

tasks/test.yaml
```diff
@@ -101,16 +101,23 @@ tasks:
...
         dir: ui
       - cmd: uds deploy k3d-core-slim-dev:latest --confirm
+      - task: deploy-metrics-server
       - task: deploy-podinfo
       - task: create-test-keycloak-user
       - task: install-runtime-in-cluster
       - cmd: npm run test:e2e-in-cluster
         dir: ui
 
+  - name: deploy-metrics-server
```
TristanHoladay, line 116: would you mind moving this to the setup file and remove the old metrics-server deploy task?

ui/src/lib/components/k8s/Drawer/component.svelte
BillyFigueroa: perfect, that makes sense

ui/src/lib/components/k8s/Logs/component.svelte
```diff
@@ -0,0 +1,277 @@
...
+              class="w-9 h-5 bg-gray-700 peer-focus:outline-none peer-focus:ring-4 peer-focus:ring-blue-800 rounded-full peer peer-checked:after:translate-x-full peer-checked:after:border-white after:content-[''] after:absolute after:top-[2px] after:left-[2px] after:bg-white after:border-gray-300 after:border after:rounded-full after:h-4 after:w-4 after:transition-all peer-checked:bg-blue-600"
+            />
+            <span class="ml-2 text-sm font-medium text-gray-300">Auto-scroll</span>
+          </label>
+        </div>
+      </div>
+
+      <!-- Right section: Action buttons -->
+      <div class="flex items-center space-x-2">
+        <div class="relative">
```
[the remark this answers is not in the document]
BillyFigueroa: I know we will need it for the apps widget since it will change from the "open" button to an icon and also, if we add a "x" one for the logs view, that will be 4 instances

I am ok with doing it in a follow up and even doing it my self

ui/src/routes/+layout.svelte
```diff
@@ -56,7 +56,7 @@
     ? 'md:ml-64'
     : 'md:ml-16'}"
 >
-  <div class="flex-grow {currRoute !== '/' ? 'overflow-hidden' : ''} p-4 pt-6">
+  <div class="flex flex-grow flex-col {currRoute !== '/' ? 'overflow-hidden' : ''} p-4 pt-6">
```
BillyFigueroa, line 59: Do we need to specify `flex-col` here since the `pods/logs/+page.svelte` file already specifies this?

ui/src/lib/components/k8s/DataTable/component.svelte
```diff
@@ -271,7 +271,11 @@
                 <tr
                   id={row.resource.metadata?.uid}
                   on:click={() =>
-                    !disableRowClick && row.resource.metadata?.uid && goto(`${baseURL}/${row.resource.metadata?.uid}`)}
+                    !disableRowClick &&
```
BillyFigueroa, line 274: it might be time to make this a function lol

ui/src/lib/components/k8s/Drawer/component.svelte
```diff
@@ -133,12 +133,21 @@
...
-
     <div class="flex-grow overflow-y-auto dark:text-gray-300 pb-20">
-      {#if activeTab === 'metadata'}
-        <!-- Metadata tab -->
+      {#if activeTab === 'details'}
+        <!-- Details tab -->
         <div class="bg-gray-800 text-gray-200 p-6 rounded-lg">
-          <dl class="space-y-4">
+          <!-- show a Logs button -->
+          <button
```
TristanHoladay, line 141: We need a mechanism to turn this off for resources without logs

Screenshot from 2024-11-06 11-48-56
Screenshot from 2024-11-06 11-49-10

tasks/test.yaml
UncleGedd: I think I can remove it entirely, I ended up using Pepr for testing as opposed to metrics server

ui/src/lib/components/k8s/Logs/Dropdown/component.svelte
```diff
@@ -0,0 +1,71 @@
...
+    dropdown.hide()
+  }
+
+  // Helper to determine what text to display
+  $: displayText = selectedValue || placeholder
+  $: isPlaceholderShown = !selectedValue
+</script>
+
+<div class="relative">
+  <span class="absolute -top-2 left-1 px-1 text-xs text-gray-400 bg-gray-800 rounded-md">
```
TristanHoladay, line 32: is this a cool tailwind thing i haven't seen or just a typo `-top-2` :laughing:

BillyFigueroa: > is this a cool tailwind thing i haven't seen or just a typo `-top-2` 😆

negative positioning :) you can put minus in most position based classes to go backwards

ui/src/lib/components/k8s/Drawer/component.svelte
UncleGedd: oof, good catch

ui/src/lib/types.ts
```diff
@@ -39,3 +39,17 @@ export type ClusterOverviewUDSPackageType = {
     endpoints: string[]
   }
 }
+
+// types for pod data loaded for the logs view
+type Container = {
+  name: string
+}
+export type Pod = {
```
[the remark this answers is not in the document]
UncleGedd: fixed

ui/src/routes/+layout.svelte
UncleGedd: I think you're right, I'll remove and check

decleaver:
Feature request? Might be because im so used to VSCode but I really wanted to be able to drag the logs view up to make it bigger lol. Not sure on the best design approach for this, but some mechanism to make this view bigger when desired would be nice.

decleaver:
nit: add some padding to the bottom of the logs component. As is, isn't as clear as to where the component ends 
image

ui/src/routes/+layout.svelte
UncleGedd: Added it back, the layout is jacked without it

ui/src/lib/components/k8s/Drawer/component.svelte
UncleGedd: fixed in latest commit

UncleGedd:
> Feature request? Might be because im so used to VSCode but I really wanted to be able to drag the logs view up to make it bigger lol. Not sure on the best design approach for this, but some mechanism to make this view bigger when desired would be nice. (This also might already be tracked somewhere for a later iteration?)

tracking here! https://github.com/defenseunicorns/uds-runtime/issues/532

decleaver:
maybe another logs view improvement? The horizontal scroll in the drop down feels a little weird. We have enough room, maybe we can make those dropdowns wider or tool tip on hover to see entire name. <insert appropriate design implementation here>
image

UncleGedd:
> nit: add some padding to the bottom of the logs component?

Added some space @decleaver 
Screenshot 2024-11-06 at 2 24 23 PM

UncleGedd:
> maybe another logs view improvement issue task? The horizontal scroll in the drop down feels a little weird. We have enough room, maybe we can make those dropdowns wider or tool tip on hover to see entire name

For this PR, I made the dropdowns a bit wider. We can explore tooltips vs scrolling as part of a follow on

ui/src/lib/components/k8s/DataTable/component.svelte
UncleGedd: Which part in particular?

BillyFigueroa: I think we just have too much logic inside this anonymous function. It starts to become difficult to read inline

ui/src/lib/components/k8s/Logs/Dropdown/component.svelte
TristanHoladay: TIL. Thanks @BillyFigueroa

ui/src/lib/components/k8s/DataTable/component.svelte
UncleGedd: 👍 tackling in a follow up, added to https://github.com/defenseunicorns/uds-runtime/issues/532

decleaver (approved):
thanks for addressing comments, looks good!

TristanHoladay (approved):

Issue #503, opened 2024-10-28
Add Gateway tab


UncleGedd:
### Is your feature request related to a problem? Please describe

Add a table for [Istio Gateways](https://istio.io/latest/docs/reference/config/networking/gateway/) nested under the Networks tab on the sidebar. The table should show:
- Name
- Namespace
- De-duped list of hosts
- Age

PR #507, opened 2024-10-29
feat: 503 add gateway tab

BillyFigueroa:
## Description
Looking to add the gateways view. As an example, we can see what this looks like on k9s

This is also adding a `hosts` column to the table

Screenshot 2024-10-29 at 12 30 40 PM
...

## Related Issue
- #503 

## Screenshots
Screenshot 2024-10-29 at 12 30 29 PM


go.mod
```diff
@@ -19,6 +19,12 @@ require (
 	k8s.io/metrics v0.31.2
 )
 
+require (
```
UncleGedd, line 22: Any idea where these deps came from? Can we remove for this PR and let Renovate handle these?

src/pkg/api/resources/cache.go
```diff
@@ -196,6 +199,12 @@ func (c *Cache) bindNetworkResources() {
 	networkPolicyGVK := coreV1.SchemeGroupVersion.WithKind("NetworkPolicy")
 	endpointGVK := coreV1.SchemeGroupVersion.WithKind("Endpoints")
 	isitoVSGVK := schema.FromAPIVersionAndKind("networking.istio.io/v1", "VirtualService")
```
UncleGedd, line 201: not in your PR but would mind correcting the spelling on `isitoVSGVK`? should be `istioVSGVK`

ui/src/lib/utils/descriptions.ts
```diff
@@ -62,4 +62,6 @@ export const resourceDescriptions: { [key: string]: string } = {
   'Zarf Packages': 'Artifact containing a set of components and configurations for air-gapped workloads.',
   UDSPackage:
     'UDS Core service layers power the platform, maintaining security, stability, and performance to keep the cluster running smoothly.',
+  Gateways:
+    'A Gateway describes a load balancer operating at the edge of the mesh receiving incoming or outgoing HTTP/TCP connections. The specification describes a set of ports that should be exposed, the type of protocol to use, SNI configuration for the load balancer, etc.',
```
UncleGedd, line 66:
```suggestion
    'An Istio Gateway describes a load balancer operating at the edge of the mesh receiving incoming or outgoing HTTP/TCP connections. The specification describes a set of ports that should be exposed, the type of protocol to use, SNI configuration for the load balancer, etc.',
```

UncleGedd:
Great work! We also need to a details view, feel to do it in this PR or as a follow-on

src/pkg/api/handlers.go
```diff
@@ -776,6 +776,21 @@ func getVirtualService(cache *resources.Cache) func(w http.ResponseWriter, r *ht
 	return rest.BindCustomResource(cache.VirtualServices, cache)
 }
 
+// @Description Get Gateways
+// @Tags gateways
```
TristanHoladay, line 780:
```suggestion
// @Tags networks
```
this will place the route under the networks section in the swagger docs

src/pkg/api/handlers.go
```diff
@@ -776,6 +776,21 @@ func getVirtualService(cache *resources.Cache) func(w http.ResponseWriter, r *ht
...
+// @Accept  html
+// @Produce text/event-stream,json
+// @Success 200
+// @Router /api/v1/resources/networks/gateways/{id} [get]
```
TristanHoladay, line 784:
```suggestion
// @Router /api/v1/resources/networks/gateways [get]
```
tested and as is it fails in swagger because of the {uid}

src/pkg/api/start.go
```diff
@@ -167,6 +167,9 @@ func Setup(assets *embed.FS) (*chi.Mux, bool, error) {
 
 				r.Get("/virtualservices", withLatestCache(k8sSession, getVirtualServices))
 				r.Get("/virtualservices/{uid}", withLatestCache(k8sSession, getVirtualService))
+
+				r.Get("/gateways", withLatestCache(k8sSession, getGateways))
+				r.Get("/gateways/{uid}", withLatestCache(k8sSession, getGateways))
```
TristanHoladay, line 172: using `getGateways()` here should work just fine for actually getting with `{uid}` but i think we have the single resource handlers (e.g. `getPod()`) for swagger docs and for each route to work when ran in swagger. Seems that if the swagger annotations mention the `{uid}` in the `@Route` then it won't work without sending a uid.

my test with Gateways in my cluster:
Screenshot from 2024-10-30 06-41-36

ui/src/lib/features/k8s/networks/gateways/store.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright 2024 Defense Unicorns
+// SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+import { ResourceStore, transformResource } from '$features/k8s/store'
+import { type ColumnWrapper, type CommonRow, type ResourceStoreInterface } from '$features/k8s/types'
+import type { VirtualService } from 'uds-core-types/src/pepr/operator/crd/generated/istio/virtualservice-v1beta1'
```
TristanHoladay, line 6: Does the `VirtualService` type also work for `Gateways`? If so, we should at least make a Gateway type that extends this one and then have Resource extend that, i think just so that it's more clear what's happening.

A little more work, but i think it would be better, is to generate the Gateway type using kubernete-fluent-client and then store that type file in our repo. But i'll defer to the rest of the team on whether that's worth it or not.

go.mod
BillyFigueroa: I am not sure. I needed to go install for swag and maybe it was added then. I ll remove

src/pkg/api/resources/cache.go
BillyFigueroa: You got it

ui/src/lib/features/k8s/networks/gateways/store.ts
BillyFigueroa: `VirtualService` does not have the field `servers`. When trying to iterate over them, typescript was not happy. This logic is adding the servers field to a newly created "Resource" via extending `VirtualService`. I called it Resource to stick to convention so we are still passing "Resource" into `transformResource` and `ResourceStore` for search-ability

src/pkg/api/start.go
BillyFigueroa: Thanks, pushed up

ui/src/lib/features/k8s/networks/gateways/store.ts
TristanHoladay: cool that makes sense. It's weird to me though to see VirtualService in the Gateways directory.

personal preference, if we use the VS type as the base, is to do something like

```tyepscript
type Gateway = VirtualService & {
  spec: VirtualService['spec'] & {
    servers: ServerType[]
  }
}

interface Resource extends Gateway {}
```

TristanHoladay (approved):

Issue #508, opened 2024-10-30
Investigate Passing `ctx` to API test Go Functions


TristanHoladay:
It could be that we don't actually need to do this, or if we do then we might want to figure out using the context in the go func().

https://github.com/defenseunicorns/uds-runtime/blob/3034fc944bcd0a97a73f101972affe91c9ec68e9/src/test/e2e/api_test.go#L84

TristanHoladay:
notes from pairing with @catsby:

1. for SSE streams we need the `go func()()` so keep that, but we don't need to pass `ctx` to the func.
1. we can get rid of the context entirely if we want and use a `time.Sleep()`
1. for non-SSE calls we don't need the `go func()()`

catsby:
I opened https://github.com/defenseunicorns/uds-runtime/pull/513 for this

PR #513, opened 2024-10-30
chore: remove contexts, use sleeps

catsby:
## Description

Minor, opinionated cleanup to e2e/api_test.go file. 

- remove context where it wasn't needed; most of the time it's purpose was to just give the background go-routine enough time to run the `ServeHTTP()` and produce a result. Instead just sleep that same time, it's less code and simpler 
- remove the go-routine that runs the http to produce a result when possible; there are times it's just one-and-done and we don't need a go-routine

## Related Issue

- #508


decleaver (approved):
cleans things up. I like it.

PR #517, opened 2024-10-30
chore: cluster overview update to use single stream

decleaver:
## Description

Updates the cluster overview page to just use one event stream vs the three it was previously using


src/pkg/api/monitor/cluster.go
```diff
@@ -101,11 +106,59 @@ func BindClusterOverviewHandler(cache *resources.Cache) func(w http.ResponseWrit
...
+			var data []unstructured.Unstructured = cache.UDSPackages.GetResources("", "")
+			var fieldsList = udsPackagesFieldsList
+
+			var filteredData = rest.FilterItemsByFields(data, fieldsList)
+
+			clusterData.PackagesData = filteredData
+			// Convert the data to JSON
+			payload, err := json.Marshal(clusterData)
+			if err != nil {
+				fmt.Fprintf(w, "data: Error: %v\n\n", err)
```
TristanHoladay, line 124: still a little unsure myself, but should these `fmt` statements be slog?

TristanHoladay (approved):

src/pkg/api/monitor/cluster.go
```diff
@@ -16,15 +16,20 @@ import (
...
-	CurrentUsage    resources.Usage   `json:"currentUsage"`
-	HistoricalUsage []resources.Usage `json:"historicalUsage"`
+type ClusterOverviewData struct {
+	TotalPods       int                      `json:"totalPods"`
+	TotalNodes      int                      `json:"totalNodes"`
+	CPUCapacity     float64                  `json:"cpuCapacity"`
+	MemoryCapacity  float64                  `json:"memoryCapacity"`
+	CurrentUsage    resources.Usage          `json:"currentUsage"`
+	HistoricalUsage []resources.Usage        `json:"historicalUsage"`
+	PackagesData    []map[string]interface{} `json:"packagesData"`
```
UncleGedd, line 26: Can this be an `unstructured` intstead of an `interface{}`?

src/pkg/api/monitor/cluster.go
```diff
@@ -36,76 +41,13 @@ func BindClusterOverviewHandler(cache *resources.Cache) func(w http.ResponseWrit
...
-			if err != nil {
-				fmt.Fprintf(w, "data: Error: %v\n\n", err)
-				return
-			}
-
-			// Write the data to the response
-			fmt.Fprintf(w, "data: %s\n\n", data)
-		}
-
 		ticker := time.NewTicker(5 * time.Second)
```
UncleGedd, line 44: Is the `ticker` doing anything?

src/pkg/api/monitor/cluster.go
```diff
@@ -36,76 +41,13 @@ func BindClusterOverviewHandler(cache *resources.Cache) func(w http.ResponseWrit
...
 		defer ticker.Stop()
 
 		// Send the initial data
-		getUsage(cache)
+		getUsage(cache, w, flusher)
```
UncleGedd, line 48: I think we'd want that flusher on all these. There's a chance that data is buffered and not sent, I believe flush ensurs that all data is sent

UncleGedd:
I see that we have a keep-alive header, but I'm not seeing any keep-alive function. Thoughts on adding this? Does the data change often enough (ie. metrics) that we don't need it?

TristanHoladay:
> I see that we have a keep-alive header, but I'm not seeing any keep-alive function. Thoughts on adding this? Does the data change often enough (ie. metrics) that we don't need it?

we might not even need these headers at this  point (though maybe for continued compatibility with HTTP1.1). Checkout the warning on this [mozilla doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection)

src/pkg/api/monitor/cluster.go
decleaver: nope, meant to delete that. Thanks!

src/pkg/api/monitor/cluster.go
decleaver: These are interface{} because I wanted to be able to use the FileterItemsByFields function that already exists in the rest package, which returns []map[string]interface{}

ui/src/lib/features/k8s/cluster-overview/component.svelte
```diff
@@ -131,18 +131,12 @@
         myChart.update()
         onMessageCount++
       }
+    } else if (newData.packagesData != null) {
```
UncleGedd, line 134: instead of checking for keys on the `newData` object, what do you think of using named events? https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

src/pkg/api/monitor/cluster.go
```diff
@@ -16,15 +16,20 @@ import (
...
 )
 
-type ClusterData struct {
-	TotalPods       int               `json:"totalPods"`
-	TotalNodes      int               `json:"totalNodes"`
-	CPUCapacity     float64           `json:"cpuCapacity"`
-	MemoryCapacity  float64           `json:"memoryCapacity"`
-	CurrentUsage    resources.Usage   `json:"currentUsage"`
-	HistoricalUsage []resources.Usage `json:"historicalUsage"`
+type ClusterOverviewData struct {
```
UncleGedd, line 19: In combination with the named events, what do you think of adding the `events` SSE stream here as well?
Screenshot 2024-11-04 at 3 22 25 PM

ui/src/lib/features/k8s/cluster-overview/component.svelte
decleaver: lol yeah, @TristanHoladay brought that up too. I can update.

src/pkg/api/monitor/cluster.go
decleaver: lol so sounded like a good idea until I started looking at the code. For the current ClusterOverviewData we end up using the data as is inside of the ClusterOverview component and passing the data into the other child components. For the events data, the event source is getting created in the ResourceStore that gets used by both the EventsOverviewWidget table and the events view table. If you're cool with it, I think I rather do this as part of an other ticket so I can look into it some more, since I think it will take a little more refactoring

UncleGedd: no problem, thanks for looking into it!

UncleGedd:
I think we have an issue, I'm rolling out a new cluster and the cluster overview page isn't updating as expected. Two examples below:

Notice how the packages are updating in K9s, but not in the widgets:
Screenshot 2024-11-04 at 9 07 26 PM


Another example, we have metrics, but the "Unavailable" error message is still there, an the CPU + memory usages don't show up
Screenshot 2024-11-04 at 9 05 54 PM

decleaver:
> I think we have an issue, I'm rolling out a new cluster and the cluster overview page isn't updating as expected. Two examples below:
> 
> Notice how the packages are updating in K9s, but not in the widgets: Screenshot 2024-11-04 at 9 07 26 PM
> 
> Another example, we have metrics, but the "Unavailable" error message is still there, an the CPU + memory usages don't show up Screenshot 2024-11-04 at 9 05 54 PM
> 
> Noting that a page refresh corrects the data but it should update automatically

Looking into it

decleaver:
@UncleGedd the reactivity seems to be better now. Reran same scenarios you ran and everything seems to update accordingly now:
image

TristanHoladay (approved):

TristanHoladay:
> @UncleGedd the reactivity seems to be better now. Reran same scenarios you ran and everything seems to update accordingly now: 

Pulled locally and confirmed updates to the widgets worked correctly.

Issue #509, opened 2024-10-30
Remove UDS Runtime Zarf packages from release artifacts


UncleGedd:
### Describe what should be investigated or refactored

Re-thinking our releases and binary distributions, we need to :
- remove the Zarf packages from our Github release (starting with our next release)
- publish `unicorn` and `registry1` Zarf package flavors to our internal 🦄 GHCR
    - recommend looking at the code in UDS Core that was removed as an example
    - ensure the GHCR path is matches other UDS packages

PR #519, opened 2024-10-31
chore: publish unicorn and registry1 flavors of runtime

decleaver:
## Description

Removes artifacts from releases and publishes "unicorn" and "registry1" flavors of runtime

## Related Issue

- #509


TristanHoladay (dismissed):

TristanHoladay:
LGTM! only thing i thought of was maybe making the IB login an action. I'm pretty sure renovate will catch the image tag in the registry1 values file as well, but we should just be on the lookout for that in case it doesn't.

Also, noting we should maybe run nightly tests for each of the flavors.

decleaver:
> LGTM! only thing i thought of was maybe making the IB login an action. I'm pretty sure renovate will catch the image tag in the registry1 values file as well, but we should just be on the lookout for that in case it doesn't.
> 
> Also, noting we should maybe run nightly tests for each of the flavors.

Not sure about renovate.... i'll look into it some more and i can update the nightly tests too, to test both flavors

decleaver:
Added smoke tests for the different flavors. Wanted to run e2e tests, but we can run into issues where the released flavor version doesn't contain updates needed to run the current e2e tests.

UncleGedd:
For the tasks:
1) `smoke` and `smoke-flavor`
2) `install-runtime-in-cluster` and `install-flavored-runtime-in-cluster`

Instead of having flavor-specific tasks can we combine these into a single one?

zarf.yaml
```diff
@@ -25,3 +27,18 @@ components:
...
+        architecture: amd64
+    images:
+      - registry1.dso.mil/ironbank/opensource/defenseunicorns/uds/runtime:v0.8.0
+    charts:
+      - name: uds-runtime
+        localPath: ./chart
+        namespace: uds-runtime
+        version: 0.1.0
+        valuesFiles:
+          - values/registry1-values.yaml
```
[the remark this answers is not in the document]
UncleGedd: You can also DRY this up a bit with the pattern over in uds-core, [example](https://github.com/defenseunicorns/uds-core/commit/0011852dd6c8f1305e2fa0c837db45f3c1801c31)

decleaver:
> For the tasks:
> 
> 1. `smoke` and `smoke-flavor`
> 2. `install-runtime-in-cluster` and `install-flavored-runtime-in-cluster`
> 
> Instead of having flavor-specific tasks can we combine these into a single one?

As is, this lets us run the smoke test and each of the flavored smoke tests in parallel in our workflows (speeding things up). This also lets us pass down the flavor var from the top level in the workflow without having to set that var in additional tasks (https://github.com/defenseunicorns/uds-runtime/pull/519/commits/a79696e69d861629082851c46a742477d63ce7ce). More than happy to combine those tasks though if preferred.

decleaver:
> > For the tasks:
> > 
> > 1. `smoke` and `smoke-flavor`
> > 2. `install-runtime-in-cluster` and `install-flavored-runtime-in-cluster`
> > 
> > Instead of having flavor-specific tasks can we combine these into a single one?
> 
> As is, this lets us run the smoke test and each of the flavored smoke tests in parallel in our workflows (speeding things up). This also lets us pass down the flavor var from the top level in the workflow without having to set that var in additional tasks ([a79696e](https://github.com/defenseunicorns/uds-runtime/commit/a79696e69d861629082851c46a742477d63ce7ce)). More than happy to combine those tasks though if preferred.

updated based on convo, adding new workflow that runs once a week or manually as desired. Refactored configuration with common zarfconfig.

.github/workflows/flavor-tests.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+# This workflow runs smoke tests for the unicorn and registry1 flavors of UDS Runtime.
+name: Flavor Tests
```
UncleGedd, line 5: How about post-release-tests or something to that effect? IMO `Flavor Tests` doesn't indicate that these are released artifacts (update file name as well)

hack/flavors/common/zarf.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+kind: ZarfPackageConfig
```
UncleGedd, line 3:
```suggestion
# Copyright 2024 Defense Unicorns
# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial

kind: ZarfPackageConfig
```

UncleGedd (dismissed):
Tested out the Zarf pkg locally, also asked @TristanHoladay to double check the YAML to try to avoid any release-day thrash

TristanHoladay (approved):

PR #522, opened 2024-11-01
fix(ui): url updates from link clicks not showing current route in sidenav

### LINKED ISSUE #487: Side Nav not showing current path after click "View Events"
If you click the "View Events" link on the Cluster Overview Events widget then you are property routed to the events table, but the side nav does not update to reflect where you're at.

Current State:
Screenshot from 2024-10-22 19-26-23

What it should be:
Screenshot from 2024-10-22 19-26-37

TristanHoladay:
## Description
Highlights the current route in the sidenav after clicking button or card that is a link. The problem was that if the link was to a route  that is a child route (e.g. /workloads/pods) and the workloads dropdown wasn't open then, even though active, the child route remained hidden.

This fix reactively looks at `$page.url.pathname` to open the correct dropdown if needed, and thus show the child route. The function is extracted so that the only variable svelte is looking at when deciding to trigger the update logic is `$page.url.pathname`. Attempts at making the entire function reactive resulted in the update logic firing when the dropdown button was toggled, meaning the submenu would be toggled off and then immediately back on.

## Related Issue

resolves #487 

[Screencast from 11-01-2024 01:59:22 PM.webm](https://github.com/user-attachments/assets/9bd5a8a8-ff6e-4b31-b562-ed98352cd8b6)


decleaver (approved):
looks good. pulled down and tested locally

Issue #515, opened 2024-10-30
Istio Gateway Details View


UncleGedd:
### Is your feature request related to a problem? Please describe

On the Gateways page, need to add a details page so that when a user clicks on rows we show the details pane

PR #523, opened 2024-11-04
feat(ui): adding ability to view drawer on click for gateway table

BillyFigueroa:
## Description
Adding ability to click on Network Istio Gateways table to open the drawer detail view

## Related Issue

- #515


UncleGedd (commented):
can we add this behavior to our nav tests?

ui/src/lib/features/k8s/networks/gateways/store.ts
```diff
@@ -24,7 +24,7 @@ interface Resource extends VirtualService {
 export type Columns = ColumnWrapper<Row>
 
 export function createStore(): ResourceStoreInterface<Resource, Row> {
-  const url = `/api/v1/resources/networks/gateways?fields=.metadata.name,.metadata.namespace,.metadata.creationTimestamp,.spec.servers`
+  const url = `/api/v1/resources/networks/gateways?fields=.metadata.name,.metadata.namespace,.metadata.uid, .metadata.creationTimestamp,.spec.servers`
```
TristanHoladay, line 27: since this list is getting a little long, could we switch to just getting all metadata?

```suggestion
  const url = `/api/v1/resources/networks/gateways?fields=.metadata,.spec.servers`
```

BillyFigueroa: > since this list is getting a little long, could we switch to just getting all metadata? we'll get a little bit of data back (labels and annotations) that we don't care about, but shouldn't cause any issues.

Updated

BillyFigueroa:
> can we add this behavior to our nav tests?

We have a drawer.spec.ts file which I believe checks the drawer. Is there a certain flow you are asking to add?

UncleGedd (approved):
Making a follow up to add tests for the Gateway details view and other details views: https://github.com/defenseunicorns/uds-runtime/issues/525

PR #526, opened 2024-11-05
chore: push go binary to s3 bucket 

decleaver:
## Description
When we cut a release we want to push the go binary to an s3 bucket that is public read-only so that it can get pulled in by the runtime CI in ironbank.


.github/workflows/tag-and-release.yaml
```diff
@@ -65,6 +65,17 @@ jobs:
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4
+        with:
+          role-to-assume: ${{ secrets.AWS_COMMERCIAL_ROLE_TO_ASSUME }}
+          role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
+          aws-region: us-west-2
+          role-duration-seconds: 21600
+
+      - name: Upload Runtime Binary to S3
+        run: aws s3 cp build/uds-runtime-linux-amd64 s3://uds-runtime/uds-runtime-linux-amd64-${{ needs.tag-new-version.outputs.release_tag }}
```
TristanHoladay, line 77: i don't think we have aws on the runner at this point. I didn't see it in our setup action and configure aws credentials says in the README it does not install the cli.

decleaver: yeah so aws cli comes preinstalled on the runners

decleaver: I didn't know that either, but there is quite a bit of software that comes pre-installed https://github.com/actions/runner-images/blob/ubuntu22/20241103.1/images/ubuntu/Ubuntu2204-Readme.md

TristanHoladay: ohhh that's awesome. not a big deal, but we could probably get rid of some of our setup in this case.

TristanHoladay (approved):

.github/workflows/tag-and-release.yaml
UncleGedd, line 77: What is the public URL for hitting this? I've tried:
- https://uds-runtime.s3.amazonaws.com/
- http://s3.amazonaws.com/uds-runtime

UncleGedd: Also can you point me to the IaC where this bucket is created?

UncleGedd: What is the URL for the binary itself? As in, the one we're going to put in the IB repo?

decleaver: haven't pushed the binary, but would be https://uds-runtime.s3.amazonaws.com/ uds-runtime-linux-amd64-<release-tag>

PR #528, opened 2024-11-05
chore: minor cleanup to avoid race

catsby:
## Description

Minor updates how we handle log streaming:

- simplify some methods, reducing goroutines which had a potential data race
- only call out for the logs once, in a consolidated call that gets the last 100 logs lines

#### Note
This PR is opened against the [logs-view PR branch](https://github.com/defenseunicorns/uds-runtime/pull/506)

#### UPDATE: 
- testing this...


catsby:
We discussed this and we're going to pull some of the changes/ideas in to the main PR and close this

PR #546, opened 2024-11-08
fix(api): pepr cache needs to be reset when cluster is switched

TristanHoladay:
## Description
There were 2 issues with the pepr streams on cluster switch:

1. the client was recreated each time with the default config overrides, which meant it was always pointed at the current context defined in the kubeconfig instead of the current context defined in the K8sSession, which is updated based on user cluster switching.

1. the pepr cache was returned to the client event if the client had been updated with the new context.

## Solution

1. turn `monitor.Pepr` into a function that returns a `http.handlerFunc` and wrap it so that we can get around closures. The wrapper had to be different than `withLatestSession` though because of import cycles that occur if we pass the entire K8sSession and then try to call `monitor.StreamCache.Reset()` in `session.Switch()`. To avoid that import cycle error, this only passes the `K8sSession.CurrentCtx` to `monitor.Pepr`, which means the function signature of the wrapper is different from `withLatestSession`.

1. we call `monitor.StreamCache.Reset()` in the `Switch()` method. I would rather the `monitor.Pepr` code keep track of some kind of context state so that it handles the logic for resetting the cache, but this is a decent enough solution for the moment. I plan to add this to the refactor list.

[Screencast from 11-08-2024 10:59:47 AM.webm](https://github.com/user-attachments/assets/bf2b15e2-72dd-4632-ade5-7ef14aec5cf5)


UncleGedd (approved):
🚀

PR #549, opened 2024-11-08
feat(ui): sprucing up details panel and logs view

UncleGedd:
## Description

- adds a number of UX improvements to the details panel and logs view (including updating logs button placement in the details panel)


decleaver (approved):
🚀

